PR #2, opened 2016-07-06
match zlib library interface and documentation where appliccable

jmoiron:
This is mostly a big documentation update, though it also introduces a bunch of breaking changes to the API to make it more similar to the stdlib's compress/zlib:

* DefaultCompressionLevel renamed to DefaultCompression
* NewWriter, NewWriterLevel, and NewWriterLevelDict introduced instead of simply NewWriter
* NewReader, NewReaderDict introduced instead of simply NewReader

Changes that shouldn't be breaking are introducing a new ErrorCode which is smaller and should be faster (though it doesn't matter) and introducing BestSpeed and BestCompression set to 1 and 20, respectively.

I've laid off the zstd C code for now, but we should change it to the new format before we release the library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,13 @@ This version has been tested and used in Datadog production environment and is s
 
 ## Usage
 
-There are two main API: simple Compress/Decompress and a streaming API (reader/writer)
+There are two main APIs: 
+
+* simple Compress/Decompress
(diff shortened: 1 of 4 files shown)


calebdoxsey:
👍

Issue #4, opened 2016-08-31
Zstandard 1.0 released


jadbox:
Has anyone tried the official 1.0 with the Go Bindings?

https://code.facebook.com/posts/1658392934479273/smaller-and-faster-data-compression-with-zstandard/

joshcarter:
FWIW I attempted to replace the .c/.h files in this package from the ones in zStd 1.0's lib folder. After fixing a header file name, I'm getting these errors:

```
could not determine kind of name for C.ZSTD_compressBegin
could not determine kind of name for C.ZSTD_compressBegin_usingDict
could not determine kind of name for C.ZSTD_compressContinue
could not determine kind of name for C.ZSTD_compressEnd
```

I did some googling around for this type of error, but didn't find any obvious fix.

Viq111:
There are a few API changes in 1.0.
Bumping up zstd version 1.0 is under way but we also need to find a way to update the package seamlessly without breaking it for people already using it (i.e: still compress with 0.5 compatibility for a bit but being able to decompress v0.5 through 1.0).

@joshcarter For `could not determine kind of name`, there is a "hidden" `ifdef` in [zstd.h](https://github.com/facebook/zstd/blob/master/lib/zstd.h#L275) that you need to set.

dgryski:
I think creating a `DataDog/zstd1` repo for v1+ is justifiable.  It's ugly, but solves the problem of backwards compatibility.

joshcarter:
@Viq111 Thanks for the tip, I set ZSTD_STATIC_LINKING_ONLY so those function definitions would be visible. I'm able to compile against the zstd shared library but I get some unit test failures:

```
    zstd_stream_test.go:22: Compressed 130000 -> 33 bytes
    zstd_stream_test.go:11: Failed to decompress with Decompress(): tableLog requires too much memory
--- FAIL: TestErrorCode (0.00s)
    zstd_test.go:48: C function returned error for -19 but ours did not
--- FAIL: TestCompressDecompress (0.00s)
    zstd_test.go:68: Compressed: [40 181 47 253 32 12 97 0 0 72 101 108 108 111 32 87 111 114 108 100 33]
    zstd_test.go:76: Error while decompressing: tableLog requires too much memory
--- FAIL: TestTooSmall (0.00s)
    zstd_test.go:108: Failed decompressing: tableLog requires too much memory
```

(I'll poke at this some more.)

[...]

PR #5, opened 2016-09-06
Updating to use zStd library version 1.0.

joshcarter:
- Removing all .c/.h files from distribution. This will link against
  the shared library instead.
- Updates to error code list. NOTE: the Go wrapper should really not
  be using its own error code map, as the upstream has no qualms about
  changing their error map at any time.
- Updated Decompress() so that it uses zStd to guess the destintion
  buffer size. If zStd doesn't provide an estimate (returns 0) then
  the previous fallback code still operates as it did before.


zstd.go
```diff
@@ -156,16 +138,31 @@ func Decompress(dst, src []byte) ([]byte, error) {
 	}
 
 	if dst == nil {
-		// x is the 95 percentile compression ratio of zstd on points.mlti payloads
-		dst = make([]byte, len(src)*3)
+		// Attempt to use zStd to determine decompressed size (may result in error or 0)
+		cSize := C.size_t(C.ZSTD_getDecompressedSize(unsafe.Pointer(&src[0]), C.size_t(len(src))))
+
+		if C.ZSTD_isError(cSize) != C.uint(0) {
```
Viq111: Calling `C.*` from Go has a perf-impact. Since we can stay in Go land, let's use the same behavior as l.76 instead of C call

zstd.go
```diff
@@ -156,16 +138,31 @@ func Decompress(dst, src []byte) ([]byte, error) {
...
+			msg := C.ZSTD_getErrorName(cSize)
+			return dst, fmt.Errorf("cannot determine decompressed size: %s", C.GoString(msg))
```
Viq111: Can we just return the `ErrorCode ` here ? (It's also already has the `String()` method)

zstd.go
```diff
@@ -156,16 +138,31 @@ func Decompress(dst, src []byte) ([]byte, error) {
 	}
 
 	if dst == nil {
-		// x is the 95 percentile compression ratio of zstd on points.mlti payloads
-		dst = make([]byte, len(src)*3)
+		// Attempt to use zStd to determine decompressed size (may result in error or 0)
+		cSize := C.size_t(C.ZSTD_getDecompressedSize(unsafe.Pointer(&src[0]), C.size_t(len(src))))
```
Viq111: I would directly come back to go-land: `size := int(...)`

zstd.go
```diff
@@ -156,16 +138,31 @@ func Decompress(dst, src []byte) ([]byte, error) {
...
+			msg := C.ZSTD_getErrorName(cSize)
+			return dst, fmt.Errorf("cannot determine decompressed size: %s", C.GoString(msg))
+		}
+
+		dstSize := int(cSize)
+
+		if dstSize > 0 {
+			dst = make([]byte, dstSize)
+		} else {
+			// x is the 95 percentile compression ratio of zstd on points.mlti payloads
```
Viq111: You can pop that comment

zstd.go
```diff
@@ -156,16 +138,31 @@ func Decompress(dst, src []byte) ([]byte, error) {
...
+			dst = make([]byte, len(src)*3)
+		}
 	}
 	for i := 0; i < 3; i++ { // 3 tries to allocate a bigger buffer
 		result, err := decompress(dst, src)
-		if err != ErrDstSizeTooSmall {
+		if err == nil {
```
Viq111: Any reasons to have changed this ?

Viq111:
Hey @joshcarter. Thanks for working so quickly on this 👍 . This is looking good.
Just an extra comment for static linking:
We actually considered both dynamic and static linking but we really prefer static linking for Go:
- You can just `go get` and you are good to go to use the library
- It's easier for vendoring
- Our (and most of the others I guess) build boxes and where the app runs are different so having the build + tests pass doesn't guarantee that your binary will work on production boxes (since it might be missing the correct zstd version there).
- But more importantly, having static linking means the executable is self-sufficient: we can just roll-in a new version, canary-test, roll-back by just downloading the executable. This is super important for our operations.
  So let's stick to static linking.

After that, we can tag this as a `v1.0` (we will do the [transition](https://github.com/DataDog/zstd/issues/4#issuecomment-245092029) on the `master` branch)

Edit: Also, we should update the `ZSTD_LICENSE` & `ZSTD_PATENT`

jmoiron:
Mattn's https://github.com/mattn/go-sqlite3 lib embeds sqlite3 source to make it go gettable but also provides build tags (passable to `go get` and friends) to link dynamically against a system installed/otherwise linkable sqlite shared library.  We should go this route which is the best of both worlds.

joshcarter:
I can put the C files back in, that's no problem. I'll submit a new pull request with that and some other minor fixes (per comments above).

joshcarter:
Ok, new commits in. I added zStandard 1.0 sources and cleaned up that error handling stuff--it was left over from some of my own experiments with zstd.

One suggestion to DataDog folks: since you have data in older formats, would it make sense to have a compressed corpus, at each version you care about, so you can verify that older versions decompress correctly? (I don't have time at the moment, sorry.)

Viq111:
looks good 👍 . I have pushed your commits to the [dev](https://github.com/DataDog/zstd/tree/dev) branch.
Also just tested trying to decode v0.5 streams and I get: `failed to zstd decompress: Unknown frame descriptor`. So I guess we will need to use [zstd magic number](https://github.com/facebook/zstd/blob/4ded9e591cbed57c54fc8f7a50412af5980e23a7/lib/legacy/zstd_v05.h#L140).
I'd like to also test for perf regression before marking this as stable and released (we run benchmark internally in real-time [here](https://p.datadoghq.com/sb/yB5yjZ-971e94528a) on a couple of nodes)

jcftang:
This 1.0 binding works reasonably well for me except I don't need backwards compatibility with older formats.

Viq111:
I tested both on OSX and linux and all tests are green and there is no regression on the benchmarks as far as I tested. I pushed those commits to the [`1.x` branch](https://github.com/DataDog/zstd/tree/1.x) and made it the default. The `0.5` compatible code will live in the [`0.5.x` branch](https://github.com/DataDog/zstd/tree/0.5.x).

`master` will help us make the transition from 0.5.x to 1.0. I don't think I have time this week to work on that but hopefully this weekend or next week.

Hopefully now people will use the `1.x` branch which is the default. Thanks for working on this @joshcarter !

PR #6, opened 2016-09-22
Adding BytesWritten() to Writer.

joshcarter:
- I have a scenario where I need to know how many bytes were written
  to the compressor's back-end. Added a private counter to Writer and
  BytesWritten() method which provides this.


joshcarter:
Addendum: feel free to skip this one, I may be changing my upper-level code so it doesn't use the streaming Writer at all. Closing the pull request.

PR #20, opened 2017-11-06
Reduce memory allocations in Reader

davies:
The reader has both decompressionBuffer and dstBuffer for buffering, seems that dstBuffer is duplicated, having decompressionBuffer should be enough.

Also, srcBuffer is not necessary.

We could reduced the memory allocation for them, also avoid the memory copying.

### DIFF
--- a/zstd_stream.go
+++ b/zstd_stream.go
@@ -226,7 +226,7 @@ func (r *reader) Read(p []byte) (int, error) {
 		if err == io.EOF && r.compressionLeft == 0 {
 			return got, io.EOF
 		} else if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
-			return got, fmt.Errorf("failed to read from underlying reader: %s", err)
+			return 0, fmt.Errorf("failed to read from underlying reader: %s", err)
 		}
 		src = src[:r.compressionLeft+n]
 
--- a/zstd_stream.go
+++ b/zstd_stream.go
@@ -147,15 +147,13 @@ func (w *Writer) Close() error {
 type reader struct {
 	ctx                 *C.ZBUFF_DCtx
 	compressionBuffer   []byte
+	compressionLeft     int
 	decompressionBuffer []byte
 	dict                []byte
 	firstError          error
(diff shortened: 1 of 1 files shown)


zstd_stream.go
```diff
@@ -209,24 +207,28 @@ func (r *reader) Close() error {
...
-		n, err := io.ReadFull(reader, src)
-		if err == io.EOF {
-			break
-		} else if err != nil && err != io.ErrUnexpectedEOF {
-			return 0, fmt.Errorf("failed to read from underlying reader: %s", err)
+		n, err := io.ReadFull(reader, src[r.compressionLeft:])
+		if err == io.EOF && r.compressionLeft == 0 {
+			return got, io.EOF
+		} else if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
+			return got, fmt.Errorf("failed to read from underlying reader: %s", err)
```
jimsmart: — Is it correct that 'got' is returned here instead of 0? Is this a typo, or is this intentional?

In Go, it is idiomatic to return the zero value with a non-nil error (except when dealing with EOF).

Much cleaner handling of buffers though, good work! — it should give a small performance boost also 👍

davies: Good call, will change that to 0.

yasushi-saito:
Can we get this change in?

Viq111 (approved):
lgtm!

Thanks for the contribution and sorry for the delay.

I tested it locally against [mr](http://sun.aei.polsl.pl/~sdeor/index.php?page=silesia) and looks good:

```
# Before
ᐅ go test -v -bench BenchmarkStreamDecompression -benchtime 30s
goos: darwin
goarch: amd64
BenchmarkStreamDecompression-4   	    2000	  31254220 ns/op	 319.01 MB/s
PASS

# After
ᐅ go test -v -bench BenchmarkStreamDecompression -benchtime 30s
goos: darwin
goarch: amd64
BenchmarkStreamDecompression-4   	    2000	  24829416 ns/op	 401.56 MB/s
PASS
ok
```

PR #32, opened 2018-03-21
Remove mallocs from compress decompress

valyala:
Go allocates a copy of slice headers when they are passed to C functions.
See golang/go#24450 for details.

Avoid this by passing unsafe.Pointer as C.uintptr_t into C functions.


Viq111:
The `master` is not used on this repository currently (will update it to mirror `1.x` default branch once we want to fork to `2.x`)
This PR will be integrated into the default `1.x` with https://github.com/DataDog/zstd/pull/31

PR #35, opened 2018-08-04
Prevent panic when decompression input slice is empty

folbricht:
This returns an error instead of an `index out of range` panic when the src slice to Decompress() is empty. The function uses `&src[0]` without first confirming the length.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -94,6 +94,9 @@ func CompressLevel(dst, src []byte, level int) ([]byte, error) {
 // prevent allocation.  If it is too small, or if nil is passed, a new buffer
 // will be allocated and returned.
 func Decompress(dst, src []byte) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


Viq111 (commented):
Hi @folbricht,

Thanks a lot for contributing!
Your PR looks good, I would just suggest adding a test to ensure now (& in the future) that we have the corret behavior.

Mind adding a test for `Decompress` on empty silce ? (should be the same as https://github.com/DataDog/zstd/blob/8bc925a36b5cdead5b9f7bfbf7780d7cd7a21907/zstd_test.go#L87)

folbricht:
Not a problem, test added. Thanks

Viq111 (approved):
Great thanks a lot!

Will merge once Travis tests are 💚

PR #56, opened 2019-04-08
[zstd] Properly keep alive src

Viq111:
In Go 1.12, inlining got more aggressive so here it can now detect that src is not used and GC earlier. We need to force tell Go that we are still using the variable.

See https://github.com/DataDog/zstd/pull/54

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"errors"
 	"io/ioutil"
+	"runtime"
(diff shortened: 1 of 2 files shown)


jmoiron (approved):
I agree with comments elsewhere that this allocation is not critical enough to make up for going off piste with conventional cgo usage, but since it's there already people may be relying on it and I don't think we should remove it.

Issue #22, opened 2018-01-15
Multiple calls to Write() has unexpected overheads?


jimsmart:
If I make multiple calls to Write, the resulting compressed data stream length is much greater than buffering the data and making a single call to Write.

— Is this expected behaviour? The reason I ask is that if chained with other writers, one cannot make any guarantees as to how they may parcel up their data.

Code (error handling omitted):
```go
func main() {

	b := &bytes.Buffer{}
	for i := 0; i < 500; i++ {
		b.Write([]byte("Hello World! "))
	}
	data1 := b.Bytes()
	fmt.Println("data len", len(data1))

	// Compress 1
	buffer1 := &bytes.Buffer{}
	w1 := zstd.NewWriterLevel(buffer1, CompressionLevel)
	w1.Write(data1)
	w1.Close()

	fmt.Println("buffer1 len", buffer1.Len())

	// Compress 2
	buffer2 := &bytes.Buffer{}
	w2 := zstd.NewWriterLevel(buffer2, CompressionLevel)
	for i := 0; i < 500; i++ {
		w2.Write([]byte("Hello World! "))
	}
	w2.Close()

	fmt.Println("buffer2 len", buffer2.Len())
}
```

Output:
```shell
data len 6500
buffer1 len 33
buffer2 len 5014
```

Regards

jimsmart:
Possibly related?
https://github.com/facebook/zstd/issues/206

jimsmart:
Ok, the best thing to do if one cannot ensure good-sized calls to Write, is to use a `bufio.Writer` like this:-

```go

func main() {

	b := &bytes.Buffer{}
	for i := 0; i < 500; i++ {
		b.Write([]byte("Hello World! "))
	}
	data1 := b.Bytes()
	fmt.Println("data len", len(data1))

	// Compress 1
	buffer1 := &bytes.Buffer{}
	w1 := zstd.NewWriterLevel(buffer1, CompressionLevel)
	w1.Write(data1)
	w1.Close()

	fmt.Println("Buffer1 len", buffer1.Len())

	// Compress 2
	buffer2 := &bytes.Buffer{}
	w2 := zstd.NewWriterLevel(buffer2, CompressionLevel)
	bw := bufio.NewWriter(w2) // default buffer size = 4k
	// bw := bufio.NewWriterSize(w2, 8192) // buffer size = 8k
	for i := 0; i < 500; i++ {
		bw.Write([]byte("Hello World! "))
	}
	bw.Flush()
[...]

Issue #39, opened 2018-09-03
[Unconfirmed] Data corruption after recent upgrade


mappu:
Hi,

Our application stores and loads compressed data on disk using this library. We recently upgraded our application from 0727e175 (tag v1.3.0) to aebefd9f (tag v1.3.4).

After this library upgrade, there were some complaints from users. After generating some JSON data, compressing it, storing it, then later loading it and decompressing it, the data could no longer be parsed (e.g. `json.Unmarshal: invalid character '\x00' in string literal`).

The issue occurred on (at least) both Windows and macOS, with both Go 1.9.7 and 1.10.3.

I assume it was caused by memory corruption in CGO data buffers.

Reverting this library back to tag v1.3.0 seems to have completely resolved the issue going-forward.

We're not yet able to reproduce the issue, and not all our users were affected (perhaps it's related to OS memory pressure?), but, just a heads-up that this library upgrade was implicated. Once we have an internal reproducer we may be able to bisect it (or hopefully, blame something else and disregard this entire issue).

x4m:
In WAL-G we observed some related data corruptions. Currently, we are hunting this too.
We have the unstable repro, but it requires a lot of setups, PostgreSQL, PITR, S3 etc.
I'll try testing 1.3.0, but it is very sporadic, kind of depends on the phase of the moon and whether on Mars.

Viq111:
Thanks for reporting!
I will try to add a fuzzer to see if we can uncover a bug.
In the meantime if you'd have any hint on the size of the payload, parallelism, type of data (or even better a reproducible payload), that would be great

mappu:
https://github.com/facebook/zstd/issues/1300 may be related (don't know).

x4m:
In my case I cannot reproduce the problem with less than 2 MAXGOROUTINES, compressed size does not exceed few Gb. I still suspect that there might be something broken in WAL-G, but lz4 and lzma do not fail on similar tests (but this is not a proof, actually, just a hint...)
v1.3.0 does not work for us either.
Maybe @Tinsane can add some more details.

PR #75, opened 2019-11-20
Add a banner to mention the data corruption bug

rasky:
Until https://github.com/DataDog/zstd/issues/22 is fixed, this library is using zstd in a way that can cause data corruption, as confirmed by zstd maintainer himself. I think this is critical enough that should be mentioned at the top of README and the Go community should be alerted until the bug is fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,10 @@
 The current headers and C files are from *v1.4.4* (Commit
 [10f0e699](https://github.com/facebook/zstd/releases/tag/v1.4.4)).
 
+# IMPORTANT NOTE: MEMORY CORRUPTION BUG
+
+Current versions of this library have a [serious bug](https://github.com/DataDog/zstd/issues/22#issuecomment-444317542) that can cause [data corruption](https://github.com/DataDog/zstd/issues/39). Until the bug is fixed, please use an alternative library such as https://github.com/valyala/gozstd.
+
 ## Usage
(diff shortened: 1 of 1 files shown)


arl:
IMHO this should be addressed.

We were using this library in production and had to switch to another one because of that.

We only noticed something was off because the sizes of the compressed files were, in some cases, **up to 4x bigger than normal**. Fortunately we didn't have data corruption.

The README indicates it is [stable](https://github.com/DataDog/zstd#stability---current-state-stable)

Viq111:
Hi @rasky & @arl,

Sorry I was out last week so getting to review this now.
Indeed, this is a pretty annoying bug that I have been trying to get a consistent reproducible test case for.
As mentioned in https://github.com/DataDog/zstd/issues/22#issuecomment-444515853, switching to using the `ZSTD_compressStream` zstd interface will likely fix the issue though I would really like to be able to consistently reproduce first to make 100% we fix both cases (#22 is easy to test #39 is harder)

@rasky Would you have a set of settings where you are able to reproduce the corruption bug consistently ?

Also if anyone is interested in contributing, I only have my free time to contribute to that repository so it's a bit harder for me to undertake big API changes

Sorry again for the wasted time this has caused to anyone
(Going to close that PR in favor of #77 but feel free to continue discussing here)

PR #80, opened 2020-02-04
[zstd_stream] Now use ZSTD_decompressStream C API

Viq111:
Follow up of https://github.com/DataDog/zstd/pull/79 but for the decompression side.

This:
- [x] add a test only for stream decompression
- [x] Potential pre-emptive fix for https://github.com/DataDog/zstd/pull/69 if there is any reuse of a reader after `Close()` has been called (see https://github.com/DataDog/zstd/commit/0932c165cc228f581afc5be24a264a433e275d3b)
- [x] Move Decompression API from the (now deprecated) ZBUFF_decompressInit / ZBUFF_decompressContinue API to ZSTD_decompressStream
- [x] Add some C passthrough methods to not have to allocated pointers
- [x] Switch from ZBUFF private methods to ZSTD methods and structs
- [x] Revert https://github.com/DataDog/zstd/pull/77 now that we use the recommended APIs

This PR is against https://github.com/DataDog/zstd/pull/79 branch not `1.x` to be able to merge both at the same time (and have a smaller diff)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,6 @@ There are two main APIs:
 The compress/decompress APIs mirror that of lz4, while the streaming API was
 designed to be a drop-in replacement for zlib.
 
-## Current status
-
-There is currently a potential issue with the streaming `Writer` interface. See https://github.com/DataDog/zstd/issues/22 for more details.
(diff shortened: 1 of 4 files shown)


Viq111:
@gmmeyer This is pretty similar to https://github.com/DataDog/zstd/pull/79 if you want to take a look as well!

gmmeyer (approved):

PR #82, opened 2020-05-17
add compression/decompression with dictionary

KlotzAndrew:
zstd supports using custom dictionaries for compression and
decompression, these are now exposed through `CompressLevelDict` and
`DecompressDict`. From an API perspective we can allow a client to
easily specify the level/dictionary or use the wrapper methods

test dictionary is generated with `zstd --train ./* -o test_dictionary`,
zstd version v1.3.3


zstd_test.go
```diff
@@ -84,6 +84,38 @@ func TestCompressDecompress(t *testing.T) {
 	}
 }
 
+// Test compression with dictionary
+func TestCompressDecompressLevelDict(t *testing.T) {
+	dict, err := ioutil.ReadFile("test_dictionary")
```
Viq111, line 89: nit: can we put test data file in a `tests_data` subfolder ?

zstd_test.go
```diff
@@ -162,6 +194,25 @@ func TestRealPayload(t *testing.T) {
...
+	}
+	rein, err := DecompressDict(nil, dst, dict)
+	if err != nil {
+		t.Fatalf("Failed to decompress: %s", err)
+	}
+	if string(raw) != string(rein) {
+		t.Fatalf("compressed/decompressed payloads are not the same (lengths: %v & %v)", len(raw), len(rein))
+	}
+}
+
```
Viq111, line 215: Do you mind also adding benchmarks `BenchmarkCompressionDict` & `BenchmarkDecompressionDict` so we can catch any future performance regressions on those APIs ?

Viq111 (commented):
Thanks for your contribution! The PR generally looks good and it's a really great addition.
I added a couple of comments to polish the PR a bit more.

PS: Sorry the delay, I actually reviewed yesterday but forgot to press submit

KlotzAndrew:
going to close this for now, didn't end up getting cycles to finish it off

Issue #90, opened 2021-02-01
TestStreamCompressionDecompressionParallel frequently fails on machines with many cpus


pterjan:
### What version of Go are you using (`go version`)?

`go version go1.15.6 linux/arm64`

### What operating system and processor architecture are you using (`go env`)?
```
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/home/iurt/.cache/go-build"
GOENV="/home/iurt/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/iurt/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/iurt/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/golang"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/golang/pkg/tool/linux_arm64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build790893092=/tmp/go-build -gno-record-gcc-switches"

```

### What did you do?

Building golang-github-datadog-zstd 1.4.5-patch1 rpm package. It sometimes passes but most of the time fails.

```
$ go test -x
WORK=/tmp/go-build935713179
mkdir -p $WORK/b001/
mkdir -p $WORK/b019/
mkdir -p $WORK/b006/
mkdir -p $WORK/b010/
mkdir -p $WORK/b029/
mkdir -p $WORK/b013/
mkdir -p $WORK/b027/
mkdir -p $WORK/b012/
mkdir -p $WORK/b030/
mkdir -p $WORK/b011/
mkdir -p $WORK/b017/
mkdir -p $WORK/b057/
[...]

PR #91, opened 2021-02-07
Pass pointers to C as unsafe.Pointer, not uintptr (memory corruption)

evanj:
The unsafe package says: "Conversion of a uintptr back to Pointer is
not valid in general." This code was violating this rule, by passing
uintptr value to C, which would then interpret them as pointers. This
causes memory corruption: https://github.com/DataDog/zstd/issues/90

This change replaces all uses of uintptr with unsafe.Pointer to avoid
this memory corruption. This has the disadvantage of marking every
argument as escaping to heap. This means the buffers used to call the
zstd functions must be allocated on the heap. I suspect this should
not be a huge problem, since I would expect that high performance
code should already be managing its zstd buffers carefully.

The bug is as follows:

* In zstd_stream_test.go: payload := []byte("Hello World!") is
  marked as "does not escape to heap" (from go test -gcflags -m).
  Therefore, it is allocated on the stack.
* The test calls Writer.Write, which converts the argument to uintptr:
  srcPtr = uintptr(unsafe.Pointer(&srcData[0]))
* Writer.Write then calls Cgo: C.ZSTD_compressStream2_wrapper(...)
* The Go runtime decides the stack needs to be larger, so it copies
  it to a new location.
* (Another thread): The Go runtime decides to reuse the old stack
  location, so it replaces the "Hello World!" bytes with new data.
* (Original thread): Calls zstd, which reads the wrong bytes.

This change adds a test which nearly always crashes for me. While
investigating the other uses of uintptr, I also was able to trigger
a similar crash when calling CompressLevel, but only with:
    GODEBUG=efence=1 go test .

I also added tests for `Ctx.CompressLevel` because it was not
obviously being tested. I did not reproduce this problem with that
function, but I suspect the same bug exists, since it uses the same
pattern.

For a minimal reproduction of this bug, see:
https://github.com/evanj/cgouintptrbug

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,16 @@ jobs:
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr go test -v'
       - run: 'PAYLOAD=`pwd`/mr go test -bench .'
+  "golang-efence":
+    resource_class: xlarge
+    docker:
+      - image: circleci/golang:latest
+    steps:
+      - checkout
+      - run: 'wget https://github.com/DataDog/zstd/files/2246767/mr.zip'
+      - run: 'unzip mr.zip'
+      - run: 'go build'
+      - run: 'PAYLOAD=`pwd`/mr GODEBUG=efence=1 go test -v'
   "golang-i386":
     docker:
       - image: 32bit/ubuntu:16.04
@@ -45,4 +55,5 @@ workflows:
       - "golang-1.14"
       - "golang-1.15"
       - "golang-latest"
+      - "golang-efence"
       - "golang-i386"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,6 @@ jobs:
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr GODEBUG=efence=1 go test -v'
   "golang-i386":
-    resource_class: xlarge
     docker:
       - image: 32bit/ubuntu:16.04
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr GODEBUG=efence=1 go test -v'
   "golang-i386":
-    resource_class: large
+    resource_class: xlarge
     docker:
       - image: 32bit/ubuntu:16.04
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,7 @@ jobs:
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr GODEBUG=efence=1 go test -v'
   "golang-i386":
+    resource_class: large
     docker:
       - image: 32bit/ubuntu:16.04
     steps:
--- a/travis_test_32.sh
+++ b/travis_test_32.sh
@@ -15,5 +15,5 @@ unzip mr.zip
 
 # Build and run tests
 go build
-PAYLOAD=$(pwd)/mr go test -v
-PAYLOAD=$(pwd)/mr go test -bench .
+DISABLE_BIG_TESTS=1 PAYLOAD=$(pwd)/mr go test -v
+DISABLE_BIG_TESTS=1 PAYLOAD=$(pwd)/mr go test -bench .
--- a/zstd.go
+++ b/zstd.go
@@ -3,27 +3,12 @@ package zstd
 /*
 #define ZSTD_STATIC_LINKING_ONLY
 #include "zstd.h"
-#include "stdint.h"  // for uintptr_t
-
-// The following *_wrapper function are used for removing superflouos
-// memory allocations when calling the wrapped functions from Go code.
-// See https://github.com/golang/go/issues/24450 for details.
-
-static size_t ZSTD_compress_wrapper(uintptr_t dst, size_t maxDstSize, const uintptr_t src, size_t srcSize, int compressionLevel) {
-	return ZSTD_compress((void*)dst, maxDstSize, (const void*)src, srcSize, compressionLevel);
-}
-
-static size_t ZSTD_decompress_wrapper(uintptr_t dst, size_t maxDstSize, uintptr_t src, size_t srcSize) {
-	return ZSTD_decompress((void*)dst, maxDstSize, (const void *)src, srcSize);
-}
-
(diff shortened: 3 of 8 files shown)


evanj:
Based on this test, we should probably change the CircleCI tests to run the tests with `GODEBUG=efence=1` to check against these types of extremely subtle issues. I can make an attempt to do that change at some point later.

Viq111:
Thanks for the PR! 
The `golang-i386` test fail seems unrelated and realted to centos6 EOL instead (Nov 2020). I'll fix it in a separate PR.
For trying to reproduce in CircleCI, I didn't have much success yet. It looks like `efence=1` uses a lot of memory so we wouldn't be able to add it to all tests but we can rerun the tests with that option on but a smaller payload. However pulling your tests (and not the code change) doesn't reproduce the bug in CircleCI: https://app.circleci.com/pipelines/github/DataDog/zstd/56/workflows/2741378d-820d-4224-897e-209f301ac000/jobs/198

Viq111:
Thanks for the PR!
I tried to import the tests onto the `1.x` (without the fix) and run with `efence` on CircleCI but unfortunatly it still looks like it's hard to replicate (i.e: the tests pass without the fix: https://app.circleci.com/pipelines/github/DataDog/zstd/61/workflows/a2a5fbd9-d76e-4de8-b981-5510372bf198/jobs/213)

evanj:
Rats you are right. I swear this was crashing reliably for me! Let me investigate a bit more...

evanj:
My test for `TestCompressLevel` was incorrect; I think I screwed up pulling the version of the test into my branch. I'll fix this PR in a moment, but it does fail in CircleCI: https://app.circleci.com/pipelines/github/DataDog/zstd/64/workflows/e8bc7fb9-a336-4f30-b077-42c6da4206b4/jobs/225

The TestStreamCompressionDecompressionParallel test does *not* fail with `GODEBUG=efence=1` which ... surprises me? I get memory corruption pretty reliably when running that test by itself on my machine, but not on CircleCI. I'll investigate that a bit more ...

evanj:
Okay, I have the other test for stream compression crashing in CircleCI as well: I needed to use more goroutines for some reason. https://app.circleci.com/pipelines/github/DataDog/zstd/67/workflows/c164d97f-9db4-46d5-83cf-2c9fa37b447f/jobs/235

This bug interestingly seems to go away with `GODEBUG=efence=1`; I think because that flag ensures that memory allocations never reuse the same memory. It only marks pages as "inaccessible" for large allocations, so I need to make the stack really big to get it to crash for me on multiple systems. I'm going to update the tests in this PR with the ones that seem to be more reliable.

In summary: this bug is very tricky, and may not happen in some cases!

Viq111 (approved):
Thanks for finding this!
It's a bit unfortunate we had to walk back on some of the optimizations but it definitely make sense given the potential memory corruptions.

Let me know if you are able to get a test we can consistently reproduce on CircleCI (Ideally we'd want that before merging this PR (and if needed we can cleanup the tests like stackDepth if it turns out it cannot reproduce the bug))

Viq111:
I ran the benchmark against the same payload we run in the tests (i.e: mr from http://sun.aei.polsl.pl/~sdeor/index.php?page=silesia) which is a ~large file of 9Mb on a Macbook Pro 2019 and there is no notable change in benchmark. I'm guessing it will likely have a slight impact on very small payloads (when you need to do a lot of calls into cgo):

```
goos: darwin
goarch: amd64
ᐅ benchstat before uintptr-bug
name                    old time/op   new time/op   delta
CtxCompression-16        91.7ms ± 6%   90.9ms ± 3%    ~     (p=0.574 n=8+8)
CtxDecompression-16      11.7ms ± 1%   11.8ms ± 1%  +0.50%  (p=0.024 n=9+9)
StreamCompression-16     96.4ms ± 2%   95.3ms ± 2%    ~     (p=0.055 n=8+10)
StreamDecompression-16   14.0ms ± 2%   13.9ms ± 1%    ~     (p=0.720 n=10+9)
Compression-16           91.2ms ± 3%   90.5ms ± 3%    ~     (p=0.579 n=10+10)
Decompression-16         11.8ms ± 2%   11.8ms ± 2%    ~     (p=0.853 n=10+10)

name                    old speed     new speed     delta
CtxCompression-16       109MB/s ± 6%  110MB/s ± 3%    ~     (p=0.574 n=8+8)
CtxDecompression-16     851MB/s ± 1%  847MB/s ± 1%  -0.49%  (p=0.024 n=9+9)
StreamCompression-16    103MB/s ± 2%  105MB/s ± 2%    ~     (p=0.055 n=8+10)
StreamDecompression-16  713MB/s ± 2%  717MB/s ± 1%    ~     (p=0.720 n=10+9)
Compression-16          109MB/s ± 3%  110MB/s ± 2%    ~     (p=0.543 n=10+10)
Decompression-16        846MB/s ± 2%  847MB/s ± 2%    ~     (p=0.853 n=10+10)
```

evanj:
I believe the tests that I added in commit https://github.com/DataDog/zstd/pull/91/commits/88ff5e14221a7fe0bbfb8e7d1c1f4443f4790742 do reproduce consistently on CircleCI. I needed to increase the number of goroutines for the parallel test, then use the recursive function stack depth trick to get it to crash with `GODEBUG=efence=1`. If you want to check that I didn't screw this up again, that would be great, but I did this separately:

https://app.circleci.com/pipelines/github/DataDog/zstd/64/workflows/e8bc7fb9-a336-4f30-b077-42c6da4206b4/jobs/225
https://app.circleci.com/pipelines/github/DataDog/zstd/67/workflows/c164d97f-9db4-46d5-83cf-2c9fa37b447f/jobs/235

Viq111:
Thanks! I ran the tests also on my branch and can confirm! https://app.circleci.com/pipelines/github/DataDog/zstd/79/workflows/a3fd4930-b663-4d4c-a57a-2a9c00740a70

`go-efence` seems to catch it. Without efence, `go-1.15` failed but `go-1.14` succeeded so it seems we can't replicate 100% of the time. I'm going to import the circleci config to add the efence test to make sure we catch those in the future as well and good to go!

Viq111:
@evanj I think I fixed the remaining test problems with the last 4 commits.
Let me know if they look good to you and I can merge

evanj:
It is very annoying that this bug depends on many implementation-specific details, so is hard to replicate. These test fixes look good to me. Thanks for checking and being so persistent with this!

PR #97, opened 2021-03-05
Fix "Go pointer to Go pointer" panics

evanj:
The Cgo runtime verifies that Go never passes pointers to other Go
pointers, which is required for correct garbage collection.
Unfortunately, its checks are not perfect, and there are occasional
false positives. Our code triggers these false positives if the
slice passed to compression functions is in the same memory
allocation as Go pointers. This happened when trying to use zstd with
another package's Writer type, which has an internal buffer.

The tests added in this PR all fail with the following panic. The
fix is to ensure the expression unsafe.Pointer(&src[0]) is inside the
Cgo call, and not before. This is documented in the following issue:

https://github.com/golang/go/issues/14210#issuecomment-346402945

The remaining uses of the "var srcPtr *byte" pattern are safe: they
all pass the address of a byte slice that is allocated internally by
this library, so I believe they cannot cause this error.

Fixes the following panic:

```
panic: runtime error: cgo argument has Go pointer to Go pointer

goroutine 30 [running]:
panic(...)
  /usr/local/go/src/runtime/panic.go:969 +0x1b9
github.com/DataDog/zstd.(*ctx).CompressLevel.func1(...)
  /home/circleci/project/zstd_ctx.go:75 +0xd9
github.com/DataDog/zstd.(*ctx).CompressLevel(...)
  /home/circleci/project/zstd_ctx.go:75 +0xce
github.com/DataDog/zstd.TestCtxCompressLevelNoGoPointers.func1(...)
  /home/circleci/project/zstd_ctx_test.go:71 +0x77
github.com/DataDog/zstd.testCompressNoGoPointers(...)
  /home/circleci/project/zstd_test.go:130 +0xad
github.com/DataDog/zstd.TestCtxCompressLevelNoGoPointers(...)
  /home/circleci/project/zstd_ctx_test.go:69 +0x37
testing.tRunner(...)
  /usr/local/go/src/testing/testing.go:1123 +0xef
```

### DIFF
--- a/zstd_stream.go
+++ b/zstd_stream.go
@@ -168,7 +168,11 @@ func (w *Writer) Write(p []byte) (int, error) {
 		srcData = w.srcBuffer
 	}
 
-	// &srcData[0] is safe: it is p or w.srcBuffer but only if len() > 0 checked above
+	if len(srcData) == 0 {
+		// this is technically unnecessary: srcData is p or w.srcBuffer, and len() > 0 checked above
+		// but this ensures the code can change without dereferencing an srcData[0]
+		return 0, nil
+	}
 	C.ZSTD_compressStream2_wrapper(
 		w.resultBuffer,
 		w.ctx,
--- a/zstd_stream.go
+++ b/zstd_stream.go
@@ -168,17 +168,13 @@ func (w *Writer) Write(p []byte) (int, error) {
 		srcData = w.srcBuffer
 	}
 
-	var srcPtr *byte // Do not point anywhere, if src is empty
-	if len(srcData) > 0 {
-		srcPtr = &srcData[0]
-	}
-
+	// &srcData[0] is safe: it is p or w.srcBuffer but only if len() > 0 checked above
 	C.ZSTD_compressStream2_wrapper(
 		w.resultBuffer,
 		w.ctx,
 		unsafe.Pointer(&w.dstBuffer[0]),
 		C.size_t(len(w.dstBuffer)),
-		unsafe.Pointer(srcPtr),
+		unsafe.Pointer(&srcData[0]),
 		C.size_t(len(srcData)),
 	)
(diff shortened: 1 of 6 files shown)


evanj:
I verified on CircleCI that the tests failed without the fix: https://app.circleci.com/pipelines/github/DataDog/zstd/90/workflows/d1eb9c40-deac-4358-81f4-ff148a65940b/jobs/333

Thanks to @niamster for telling me about this! This is an interesting one. Cgo is annoying.

zstd_stream.go
```diff
@@ -168,17 +168,13 @@ func (w *Writer) Write(p []byte) (int, error) {
 		srcData = w.srcBuffer
 	}
 
-	var srcPtr *byte // Do not point anywhere, if src is empty
-	if len(srcData) > 0 {
-		srcPtr = &srcData[0]
-	}
-
+	// &srcData[0] is safe: it is p or w.srcBuffer but only if len() > 0 checked above
```
Viq111, line 171: I would still do some defensive coding do an if statement like the others (but that's just personal opinion), here we assume anyone making change to the `Write` method will read that comment (i.e: if I remove `if len(p) ==0` condition above, I might not read this comment)

Viq111 (approved):
Thanks for the tests and the fix!

Would you mind making it not a draft PR anymore so we can merge it ?

I think @niamster tested it well so it should be good to go on this front as well

This should fix https://github.com/DataDog/zstd/issues/98 once merged

zstd_stream.go
evanj: This is a good point. This is also reasonably complex logic to figure out that this is safe: it involves following some complex operations above. In this case, we could do another "early return": `if len(srcData) == 0 { return 0, nil }`. I will make this change.

Viq111 (approved):

PR #104, opened 2021-06-21
Add support for linking to system libzstd

WGH-:
Inspired by https://github.com/mattn/go-sqlite3


WGH-:
Okay, apparently, it is not that simple: libzstd version included in Debian Buster is little bit old

Viq111:
Going to close this one in favor of https://github.com/DataDog/zstd/pull/109
Thanks for doing it, it definitely put @dopuskh3 on the right track

PR #105, opened 2021-07-07
zstd.go: Set ZSTD_LEGACY_SUPPORT=4 to decompress legacy payloads

evanj:
The compressed output format was finalized with zstd version 0.8. To
support decompression of previous versions, set the macro
ZSTD_LEGACY_SUPPORT=4, which matches the documented default for the
zstd command line tool:

https://github.com/facebook/zstd/blob/dev/programs/README.md

This made the compiled library about 20 kiB larger, but will allow
this to be used with really old zstd payloads.

.circleci/config.yml: Do not test with different versions of this
macro, since the tests now include a check that it can decompress
old payloads.


niamster:
By default `ZSTD` prefers to have legacy formats enabled, see discussion in upstream PR https://github.com/facebook/zstd/pull/3079
Homebrew `ZSTD` maintainer also agreed to enabled it by default, see discussion in upstream PR https://github.com/Homebrew/homebrew-core/pull/95735

Viq111 (approved):
Thanks!

PR #109, opened 2021-11-18
Add support for building zstd binding against an external libzstd library

dopuskh3:
This is heavily inspired by [the following PR](https://github.com/DataDog/zstd/pull/104/files) from @WGH-

Add the `external_libzstd` tag to be used when we want to build the binding against an externally provided 
zstd library. If the tag is defined, the go build system will look for zstd pkg-config package to gather 
build arguments.

Added CI jobs to run tests and benchmarks with latest go version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,11 +11,35 @@ jobs:
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr go test -v'
       - run: 'PAYLOAD=`pwd`/mr go test -bench .'
+  "golang-1.16-external-libzstd":
+    docker:
+      - image: circleci/golang:1.16
+    steps:
+      - checkout
+      - run: 'sudo apt update'
+      - run: 'sudo apt install libzstd-dev'
+      - run: 'wget https://github.com/DataDog/zstd/files/2246767/mr.zip'
+      - run: 'unzip mr.zip'
+      - run: 'go build'
+      - run: 'PAYLOAD=`pwd`/mr go test -v'
+      - run: 'PAYLOAD=`pwd`/mr go test -bench .'
   "golang-1.17":
     docker:
       - image: circleci/golang:1.17
     steps:
       - checkout
+      - run: 'wget https://github.com/DataDog/zstd/files/2246767/mr.zip'
+      - run: 'unzip mr.zip'
+      - run: 'go build'
+      - run: 'PAYLOAD=`pwd`/mr go test -v'
+      - run: 'PAYLOAD=`pwd`/mr go test -bench .'
+  "golang-1.17-external-libzstd":
+    docker:
+      - image: circleci/golang:1.17
+    steps:
+      - checkout
+      - run: 'sudo apt update'
+      - run: 'sudo apt install libzstd-dev'
       - run: 'wget https://github.com/DataDog/zstd/files/2246767/mr.zip'
       - run: 'unzip mr.zip'
       - run: 'go build'
@@ -86,7 +110,9 @@ workflows:
   build:
     jobs:
       - "golang-1.16"
+      - "golang-1.16-external-libzstd"
       - "golang-1.17"
+      - "golang-1.17-external-libzstd"
       - "golang-latest"
       - "golang-latest-external-libzstd"
       - "golang-efence"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,18 @@ jobs:
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr go test -v'
       - run: 'PAYLOAD=`pwd`/mr go test -bench .'
+  "golang-latest-external-libzstd":
+    docker:
+      - image: circleci/golang:latest
+    steps:
+      - checkout
+      - run: 'sudo apt update'
+      - run: 'sudo apt install libzstd-dev'
+      - run: 'wget https://github.com/DataDog/zstd/files/2246767/mr.zip'
+      - run: 'unzip mr.zip'
+      - run: 'go build -tags external_libzstd'
+      - run: 'PAYLOAD=`pwd`/mr go test -tags external_libzstd -v'
+      - run: 'PAYLOAD=`pwd`/mr go test -tags external_libzstd -bench .'
   "golang-efence":
     resource_class: xlarge
     docker:
@@ -41,6 +53,18 @@ jobs:
       - run: 'unzip mr.zip'
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr GODEBUG=efence=1 go test -v'
+  "golang-efence-external-libzstd":
+    resource_class: xlarge
+    docker:
+      - image: circleci/golang:latest
+    steps:
+      - checkout
+      - run: 'sudo apt update'
+      - run: 'sudo apt install libzstd-dev'
+      - run: 'wget https://github.com/DataDog/zstd/files/2246767/mr.zip'
+      - run: 'unzip mr.zip'
+      - run: 'go build -tags external_libzstd'
+      - run: 'PAYLOAD=`pwd`/mr GODEBUG=efence=1 go test -tags external_libzstd -v'
   "golang-zstd-legacy-support":
     docker:
       - image: circleci/golang:latest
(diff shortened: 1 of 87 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,6 @@
 version: 2
-
 jobs:
-  "golang-1.14":
+  golang-1.14:
```
evanj, line 3: It looks like this is just a formatting change, but I wonder if we should be updating these? The supported Go releases at this moment in time are Go 1.16, and Go 1.17. I'm not sure we should try to keep this working on older releases?

... although this should probably be a separate change.

external_zstd.go
```diff
@@ -0,0 +1,7 @@
+// +build external_libzstd
```
evanj, line 1: Can you run `go fmt .` on this file using Go 1.17? I believe it should add a separate line with the "new style" build directives, which look like `//go:build` ; See https://golang.org/doc/go1.17#build-lines

(this may imply that CI in this repository is missing a check? Or maybe is running checks with an older version of Go)

insert_libzstd_ifdefs.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
```
evanj, line 1: Maybe this file should include a header comment describing what it is and why it exists? E.g.=

```
"""
This script rewrites the zstd files ... . When updating zstd, do x y z.
"""
```

insert_libzstd_ifdefs.py
```diff
@@ -0,0 +1,28 @@
...
+
+    with open(file, "r") as fd:
+        content=fd.read()
+
+    with open(file, "w") as fd:
+        fd.seek(0)
+        fd.write(f"#ifndef {FLAG}\n")
+        fd.write(content)
+        if file == "zstd.h":
+            fd.write("\n".join([
```
evanj, line 17: I find this hard to read due to the line breaks. maybe it should be a top-level multi-line constant? E.g.

ZSTD_H_SUFFIX = f"""
#else ...
#endif"""
```

then this line would just be `fd.write(ZSTD_H_SUFFIX)`

File: insert_libzstd_ifdefs.py [hunk 5]
```diff
@@ -0,0 +1,28 @@
...
+
+FLAG="USE_EXTERNAL_ZSTD"
+
+def insert_ifdefs(file):
+
+    with open(file, "r") as fd:
+        content=fd.read()
+
+    with open(file, "w") as fd:
+        fd.seek(0)
```
evanj, line 13: I'm *pretty* sure this is unnecessary, since opening a file for writing and not appending should start at the beginning of the file? Doesn't really matter though ...

evanj (approved):
I think this looks good, but we should get @Viq111 to take a look.

I also wonder if this package should be renaming all the zstd symbols by default, to allow it to be linked along side other versions of zstd? I think we would *still* want the option to use an external version, but this might reduce the number of places where people would need this option? (clearly that would be another thing to do in a future change, not here)

WGH-:
When I was working on the PR, I got stuck because Debian Buster (10), on which golang Docker image was based on, had relatively old libzstd, and some of the .go files used new functions that weren't present. I couldn't find an easy way to make CI tests pass.

I suppose you circumvented this issue by waiting until golang images started to base on Debian Bullseye (11)?

If so, does it mean that `external_libzstd` won't work on older libzstd versions, even if user doesn't use any of those new functions?

.circleci/config.yml
dopuskh3: fixed

external_zstd.go
dopuskh3: done

insert_libzstd_ifdefs.py
dopuskh3: done

insert_libzstd_ifdefs.py
dopuskh3: fixed

insert_libzstd_ifdefs.py
dopuskh3: fixed

dopuskh3:
> When I was working on the PR, I got stuck because Debian Buster (10), on which golang Docker image was based on, had relatively old libzstd, and some of the .go files used new functions that weren't present. I couldn't find an easy way to make CI tests pass.
> 
> I suppose you circumvented this issue by waiting until golang images started to base on Debian Bullseye (11)?
> 
> If so, does it mean that `external_libzstd` won't work on older libzstd versions, even if user doesn't use any of those new functions?

`external_libzstd` will work as soon as the symbols used in the go bindings are available in the lib `zstd.h` located in the path returned by `pkg-config`. Unfortunately, cgo doesn't provide the ability to ask for dependency prerequisites (it could for example use `pkg-config` `--atleast/exact/max-version` feature)

dopuskh3:
> I think this looks good, but we should get @Viq111 to take a look.
> 
> I also wonder if this package should be renaming all the zstd symbols by default, to allow it to be linked along side other versions of zstd? 
I agree that bundling symbols that may be provided to the compiler by other projects linking with zstd can make the situation a little bit tricky to handle for larger projects. Specially, in this situation, forgetting to pass the `external_libzstd` tag to go build can end up silently use vendored source code for other dependencies using libzstd:

Here is what I observed in the following situation:
```
  [binary] <---- [DataDog/zstd] <---- [vendored zstd code]
     ^
     |
     +---------[confluent-kafka-go] <---- [librdkafka] <----[libzstd]
```
I observed that the calls to confluent-kafka-go using zstd actually ended up running the `[vendored zstd code]`, which can be annoying in multiple ways.

In order to avoid this situation to silently happen, we could make passing a build tag mandatory by failing the compilation intentionally if none of the `vendored_libztd` nor `external_libzstd` is passed.

.circleci/config.yml
```diff
@@ -31,6 +31,18 @@ jobs:
       - run: 'go build'
       - run: 'PAYLOAD=`pwd`/mr go test -v'
       - run: 'PAYLOAD=`pwd`/mr go test -bench .'
+  "golang-latest-external-libzstd":
```
Viq111, line 34: you can probably update Go 1.14, 1.15 with 1.16 & 1.17 (we just want to make sure we support current go version + previous one)

insert_libzstd_ifdefs.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
```
Viq111, line 1: Do you mind putting this file in a `tools/` folder ?
That way it's much easier to discover (and we can add later on a script to pull the latest zstd c upstream)

zstd.h
```diff
@@ -2530,3 +2531,7 @@ ZSTDLIB_API size_t ZSTD_insertBlock    (ZSTD_DCtx* dctx, const void* blockStart,
 #if defined (__cplusplus)
 }
 #endif
+#else /* USE_EXTERNAL_ZSTD */
+#undef ZSTD_STATIC_LINKING_ONLY
```
Viq111, line 2535: This is a bit hacky to define `ZSTD_STATIC_LINKING_ONLY` in the Go file and then unset in the C file. Is there a way to just have the switch in `external_zstd.go` ? (i.e: an else that sets it)
(I don't remember if we need to define & import `zstd.h` in each go file)

Viq111 (commented):
Thanks! This looks good to me, I just added some small comments

zstd.h
dopuskh3: Are you sure we actually need to define ZSTD_STATIC_LINKING_ONLY? It appears that library is building fine when that's not defined.

dopuskh3:
> > When I was working on the PR, I got stuck because Debian Buster (10), on which golang Docker image was based on, had relatively old libzstd, and some of the .go files used new functions that weren't present. I couldn't find an easy way to make CI tests pass.
> > I suppose you circumvented this issue by waiting until golang images started to base on Debian Bullseye (11)?
> > If so, does it mean that `external_libzstd` won't work on older libzstd versions, even if user doesn't use any of those new functions?
> 
> `external_libzstd` will work as soon as the symbols used in the go bindings are available in the lib `zstd.h` located in the path returned by `pkg-config`. Unfortunately, cgo doesn't provide the ability to ask for dependency prerequisites (it could for example use `pkg-config` `--atleast/exact/max-version` feature)

@WGH- I added version constraint with preprocessor macros in the external_libzstd.go file [here](https://github.com/DataDog/zstd/pull/109/commits/0c4231e2c451a9233062cfe376c3ab09db1daca0)

Viq111 (approved):
lgtm!

Maybe one last thing (that doesn't necessitate another review). Could you add the `go build -tags external_libzstd` as a section in the README ? That way it's discoverable for users without needing to read the code

PR #114, opened 2022-04-02
Implements Bulk processing dictionary API

Barbayar:
## Description
Implements [Bulk processing dictionary API](https://github.com/DataDog/zstd/blob/718c6ee5063f765c9bd64a10de17968e426b9c0c/zstd.h#L841).

## Bulk processing dictionary API
It allows you to compress files with a pre-trained dictionary. Because, the compress file doesn't contain a dictionary, you can save some bytes. The save is significant if your files are smaller and similar.

P.S. With [Streaming API](https://github.com/DataDog/zstd/blob/1.x/zstd_stream.go), we can compress/decompress files with a pre-trained dictionary. But, in order to compress/decompress multiple files, we need to re-create writer/reader every time we compress/decompress. It's very inefficient, because it needs to load the dictionary every time.

## Usage
```
p, err := NewBulkProcessor(dict, BestSpeed)
if err != nil {
	fmt.Fatalf("failed to create a BulkProcessor")
}

for payload := range in {
	compressed, err := p.Compress(nil, payload)
	if err != nil {
		fmt.Fatalf("failed to compress")
	}

	uncompressed, err := p.Decompress(nil, compressed)
	if err != nil {
		fmt.Fatalf("failed to decompress")
	}
}
```

### DIFF
--- a/zstd_bulk.go
+++ b/zstd_bulk.go
@@ -6,10 +6,26 @@ package zstd
 import "C"
 import (
 	"errors"
+	"runtime"
 	"unsafe"
 )
 
-// BulkProcessor implements Bulk processing dictionary API
+var (
+	// ErrEmptyDictionary is returned when the given dictionary is empty
+	ErrEmptyDictionary = errors.New("Dictionary is empty")
+	// ErrBadDictionary is returned when cannot load the given dictionary
+	ErrBadDictionary = errors.New("Cannot load dictionary")
+	// ErrContentSize is returned when cannot determine the content size
+	ErrContentSize = errors.New("Cannot determine the content size")
+)
+
+// BulkProcessor implements Bulk processing dictionary API.
+// When compressing multiple messages or blocks using the same dictionary,
+// it's recommended to digest the dictionary only once, since it's a costly operation.
+// NewBulkProcessor() will create a state from digesting a dictionary.
+// The resulting state can be used for future compression/decompression operations with very limited startup cost.
+// BulkProcessor can be created once and shared by multiple threads concurrently, since its usage is read-only.
+// The state will be freed when gc cleans up BulkProcessor.
 type BulkProcessor struct {
 	cDict *C.struct_ZSTD_CDict_s
 	dDict *C.struct_ZSTD_DDict_s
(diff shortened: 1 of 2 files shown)


zstd_bullk_test.go
```diff
@@ -0,0 +1,152 @@
...
+	wordCount := 10 + rand.Intn(100) // 10 - 109
+	result := []string{}
+	for i := 0; i < wordCount; i++ {
+		result = append(result, words[rand.Intn(len(words))])
+	}
+
+	return strings.Join(result, " ")
+}
+
+func TestCompressAndDecompress(t *testing.T) {
```
Viq111, line 30: can you prefix the tests with `TestBulk` so we don't have collisions in the future ?
https://github.com/DataDog/zstd/blob/1.x/zstd_test.go#L59
(i.e: `TestBulkCompressDecompress`)

zstd_bullk_test.go
```diff
@@ -0,0 +1,152 @@
...
+	var b64 = base64.StdEncoding
```
Viq111, line 31: nit: that var seems a bit redundant, either define it at the test package level or don't use the var ?

zstd_bullk_test.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		t.Fatalf("failed to create a BulkProcessor")
+	}
+
+	for i := 0; i < 100; i++ {
+		payload := []byte(getRandomText())
+
+		compressed, err := p.Compress(nil, payload)
+		if err != nil {
+			t.Fatalf("failed to compress")
```
Viq111, line 47: those are probably `t.Error` not `t.Fatal`.
and you can probably rename all `Fatalf` to `Fatal`

zstd_bullk_test.go
```diff
@@ -0,0 +1,152 @@
...
+	payload, err := p.Compress(nil, []byte("We're building a platform that engineers love to use. Join us, and help usher in the future."))
+	if err != nil {
+		b.Fatalf("failed to compress")
+	}
+	for n := 0; n < b.N; n++ {
+		_, err := p.Decompress(nil, payload)
+		if err != nil {
+			b.Fatalf("failed to decompress")
+		}
+		b.SetBytes(int64(len(payload)))
```
Viq111, line 148: this should probably be outside the loop

zstd_bullk_test.go
```diff
@@ -0,0 +1,152 @@
+package zstd
```
Viq111, line 1: could you maybe test corner cases (see other tests for `zstd_test.go` and `zstd_stream_test.go` and `zstd_ctx_test.go`) ? Also I think there is some general cleanup you could do to make the code clearer (I pointed some below)

Viq111 (commented):
Thanks for the PR! The structure looks straight forward, I've added a couple of comments to make it more clear and robust for an external audience

zstd_bulk.go
```diff
@@ -0,0 +1,116 @@
...
+// Compress compresses the `src` with the dictionary
+func (p *BulkProcessor) Compress(dst, src []byte) ([]byte, error) {
+	bound := CompressBound(len(src))
+	if cap(dst) >= bound {
+		dst = dst[0:bound]
+	} else {
+		dst = make([]byte, bound)
+	}
+
+	var cSrc unsafe.Pointer
```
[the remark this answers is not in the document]
Barbayar: Okay, this was new to me. Thanks for pointing out.

Viq111 (approved):
Thanks! this looks great!

Issue #108, opened 2021-11-02
ctx creation and customition


clor09:
Please answer these questions before submitting your issue. Thanks!

### What version of Go are you using (`go version`)?
go version go1.15.9 linux/amd64

### What operating system and processor architecture are you using (`go env`)?
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/xx/.cache/go-build"
GOENV="/home/xx/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/xx/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/xx/go"
GOPRIVATE=""
GOPROXY=""
GOROOT="/usr/lib/go-1.15"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go-1.15/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build605241201=/tmp/go-build -gno-record-gcc-switches"

### What did you do?

m wondering if we could export the ctx creation to the users so that they can make some customition, i have noticed the very early issue #17  and i did it myself by adding the following code to zstd_streaming.go, but maybe it's really not a good impl.
```go
#cgo CPPFLAGS: -DZSTD_MULTITHREAD
#cgo CFLAGS: -pthread

func (w *Writer) SetNbWorkers(n int) error {
	if w.firstError != nil {
		return w.firstError
	}
	if err := getError(int(C.ZSTD_CCtx_setParameter(w.ctx, C.ZSTD_c_nbWorkers, C.int(n)))); err != nil {
		w.firstError = err
		return err
	}
	return nil
}
```

### What did you expect to see?
compress&decompress ctx creation & customition avaliable

### What did you see instead?
ctx created by the lib internally and it's not accessable outside

[...]

PR #117, opened 2022-05-05
Add SetNbWorkers api to the writer code (see #108)

bsergean:


bsergean:
zstd might support parallel decompression, if this is the case maybe we could put the same api on the reader.

Viq111:
Thanks for your contribution!
Would you mind adding a test that check that the error returned is nil and maybe via `ZSTD_CCtx_getParameter` that the parameter is correctly set ?

bsergean:
Thanks for the tip, I’ll add a test, and check how ZSTD_CCtx_getParameter works.

bsergean:
```
zstd$ go test -run TestStreamSetNbWorkers
--- FAIL: TestStreamSetNbWorkers (0.00s)
    zstd_stream_test.go:407: Expected SetNbWorkers to succeed, got Unsupported parameter instead
```

Interestingly my unittest started failing as is, as the parameter was seen as unsupported. This is because the library was not built in threading mode, which I changed. I don't know if that's acceptable, but it's a requirement for this PR.

bsergean:
I'm getting that error in the unittest in the efence build -> https://app.circleci.com/pipelines/github/DataDog/zstd/179/workflows/5f082f6d-b02f-4434-981c-9bacb0297e85/jobs/916

```
    zstd_stream_test.go:19: Failed writing to compress object: Unsupported parameter
```

If I try to run efence locally, I'm getting a panic ...

```
GODEBUG=efence=1 go test -run TestStreamSetNbWorkers
```

zstd_stream_test.go
```diff
@@ -19,10 +20,14 @@ func failOnError(t *testing.T, msg string, err error) {
 	}
 }
 
-func testCompressionDecompression(t *testing.T, dict []byte, payload []byte) {
+func testCompressionDecompression(t *testing.T, dict []byte, payload []byte, nbWorkers int) {
 	var w bytes.Buffer
 	writer := NewWriterLevelDict(&w, DefaultCompression, dict)
-	_, err := writer.Write(payload)
+
+	err := writer.SetNbWorkers(nbWorkers)
```
Viq111, line 27: I would probably only call this is `nbWorkers` > 1 so you can test the (most common) path where the user doesn't call this method

zstd_stream_test.go
```diff
@@ -19,10 +20,14 @@ func failOnError(t *testing.T, msg string, err error) {
...
+	failOnError(t, "Failed writing to compress object", err)
```
Viq111, line 28: as mentioned, you might want to skip the test (for nbWorkers > 1) if the error is that the library doesn't support multi-threading

zstd_stream_test.go
```diff
@@ -398,6 +403,27 @@ func TestStreamWriteNoGoPointers(t *testing.T) {
 	})
 }
 
+func TestStreamSetNbWorkers(t *testing.T) {
+	// Build a big string first
+	sb := strings.Builder{}
```
Viq111, line 408: you are not testing the `strings.Builder{}` for a one-liner, I would just `s := strings.Repeat("foobaa", 1000*1000)`

Viq111 (commented):
Sorry for the delayed answer.

For the dynamic library not built with `-DZSTD_MULTITHREAD=1`, I think it's fair to assume that we may encounter it from time to time in the wild a library.
I think it's fine but we should make the error more user friendly, what about introducing an error: `ErrNoParallelSupport` and return it when the caller uses `SetNbWorkers` but the underlying library doesn't support it ?
In tests you can then do a:
```go
if err := writer.SetNbWorkers(nbWorkers); err == ErrNoParallelSupport {
  t.Skip()
}
```

What do you think ?

bsergean:
No worries, thanks for the feedback. Before starting to implement the changes I want to clarify one thing:

I didn’t know that a shared library mode was supported, I thought that all the C files in zstd were built with the cgo module. Doesn’t the commit below imply that we build all the C code ?
```
commit af19ae6bd5af997c006a9e6dfcbf24a211d24e18
Author: Josh Carter ***@***.***>
Date:   Wed Sep 7 12:24:53 2016 -0600

    Adding C sources for zStd 1.0 in favor of shared lib.
```

Viq111:
This [PR](https://github.com/DataDog/zstd/pull/109) added support to either compile with the dynamic library or static library (added in the README [here](https://github.com/DataDog/zstd#building-against-an-external-libzstd)).
The default (no compile flags) is to build statically

bsergean:
Got it, thanks for digging that PR. 

I’ll try to implement the change you suggested, and might poke the PR if I fail to do so.

zstd_stream_test.go
bsergean: I think this is addressed.

zstd_stream_test.go
bsergean: I think this is addressed.

bsergean:
I introduced the new error type you suggested (made it global by giving it a starting capital letter).

All the tests are passing now, I wonder if there was a fluke previously, or if one env has an external library that was built without parallel support, which would be great.

Lastly, I have not tested that code very much, to measure the performance impact (I know we use multiple threads, and have to use C to do that).

Viq111 (approved):
Thanks!
I will do some benchmarking on different hardware and let you know but PR looks good!

bsergean:
Ah excellent. I’ve been too lazy to do the benchmarking myself, I’ll try to do it but I can’t promise anything.

Viq111:
## 

## edit: this comment is incorrect, [see latest](https://github.com/DataDog/zstd/pull/117#issuecomment-1147812767)


I have been benchmarking with the [mozilla](https://sun.aei.polsl.pl//~sdeor/corpus/mozilla.bz2) (unbzip2 then tarred) payload (the biggest one).

On x86, the performance is always better, albeit much more variable.
Without parallelism:
```
goos: darwin
goarch: amd64
pkg: github.com/DataDog/zstd
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkStreamCompression-12    	      85	 392153074 ns/op	 135.01 MB/s
BenchmarkStreamCompression-12    	      82	 389099060 ns/op	 136.07 MB/s
BenchmarkStreamCompression-12    	      86	 396593782 ns/op	 133.50 MB/s
BenchmarkStreamCompression-12    	      88	 387180328 ns/op	 136.74 MB/s
BenchmarkStreamCompression-12    	      88	 388614650 ns/op	 136.24 MB/s
```

With parallelism = 8:
```
goos: darwin
goarch: amd64
pkg: github.com/DataDog/zstd
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkStreamCompression-12    	     100	 323739454 ns/op	 163.54 MB/s
BenchmarkStreamCompression-12    	     132	 260639012 ns/op	 203.13 MB/s
BenchmarkStreamCompression-12    	     130	 266204210 ns/op	 198.89 MB/s
BenchmarkStreamCompression-12    	     139	 263071863 ns/op	 201.25 MB/s
BenchmarkStreamCompression-12    	     159	 442939701 ns/op	 119.53 MB/s
```

Summary:
```
name                  old time/op   new time/op   delta
StreamCompression-12    391ms ± 2%    311ms ±42%   ~     (p=0.151 n=5+5)

name                  old speed     new speed     delta
StreamCompression-12  136MB/s ± 1%  177MB/s ±33%   ~     (p=0.151 n=5+5)
```

Howerver on ARM (testing on a `c7g.4xlarge`), with parellization=8 on 16 cores, I unfortunately get OOMs so looking into it:
```
goos: linux
goarch: arm64
pkg: github.com/DataDog/zstd
BenchmarkStreamCompression-16    	Killed
```

bsergean:
In general I expect multi-threaded work to use more memory, maybe your arm test machine has less RAM than your Intel test machine ?

I wonder if the zstd cli has a nb worker option, this could be used also to debug this.

Viq111:
Yeah same expectation here but the mozilla payload is ~50Mb and with 8 concurrent threads, I would not expect it to starve the 32GiB of memory a `c7g.4xlarge`have

bsergean:
BTW I have an apple M1 so I should be able to reproduce the problem possibly. I think my Mac actually has 32G of ram.

Viq111:
👍 sounds good, here are my current testing steps:

Add `w.SetNbWorkers(8)` to the [benchmark](https://github.com/DataDog/zstd/blob/1.x/zstd_stream_test.go#L406)

then
```
wget https://sun.aei.polsl.pl//~sdeor/corpus/mozilla.bz2
tar -xf mozilla.bz2 mozilla
tar cf mozilla.tar mozilla
export PAYLOAD=`pwd`/mozilla.tar
go test -c
./zstd.test -test.run None -test.bench BenchmarkStreamCompression -test.benchtime 30s -test.count 5
```

On a `c7.4xlarge`, you can see it uses all (32Gib) of the memory:
```
$ grep ^VmPeak /proc/$(pidof zstd.test)/status
VmPeak:	33758272 kB
```

bsergean:
I think I can reproduce the high memory usage pb:

zstd.test memory usage went as high as 28Gig, I interrupted the test because I didn’t want my main laptop to die ...

bsergean:
With only one worker, the peak memory usage is 148Mib, so something odd is going on.

>>> 30000 / 148.
202.7027027027027

200 is the max number of workers that is set by zstd. Maybe it isn’t a coincidence ?

Viq111:
On a `c6i.4xlarge` (16 vCPUs, 32GiB, x86), the peak memory usage for the same scenario is much more sane:
```
grep ^VmPeak /proc/$(pidof zstd.test)/status
VmPeak:	 1531488 kB
```
which is 1.46GiB

I will try to reproduce with the CLI and if we see the same, I think it's fair to merge this PR as it lies on the C side

Viq111:
Sorry for my previous message, on a `c6i.4xlarge` (x86) I have the same consumption:
```
grep ^VmPeak /proc/$(pidof zstd.test)/status
VmPeak:	33356928 kB
```
so going to look on the side of the C code indeed

Viq111:
Ok I think I found the "issue".
See https://github.com/DataDog/zstd/commit/9dd8a8a69ed8c30a736bbc8694ffa2bb0eeadbcc
A gotcha was when setting [nbWorkers >1](https://github.com/facebook/zstd/blob/eadb6c874f9d0c9e90c835f8b0181da802361e4c/doc/zstd_manual.html#L318-L321), the compress C calls becomes asynchronous instead of synchronous. That means with Go benchmark trying to do benchmarks with an increasing `N` number, we would buffer a lot of data on the C side.

By doing https://github.com/DataDog/zstd/commit/9dd8a8a69ed8c30a736bbc8694ffa2bb0eeadbcc and forcing a flush every 1GiB, we force the C zstd buffer to not have more than 1GiB at a time.

This works:
```
grep ^VmPeak /proc/12092/status                                                                                                                                                                                           Mon Jun  6 18:54:05 2022

VmPeak:  8307060 kB
```
Memory usage is 8GiB which is exactly 1GiB for 8 workers.

So there is nothing wrong with the code, just the benchmark did not take the async into account.

The good news is the performance with 8 workers is much better than 1 worker:
```
name                  old time/op   new time/op   delta
StreamCompression-16    414ms ± 1%     98ms ± 2%   -76.37%  (p=0.008 n=5+5)

name                  old speed     new speed     delta
StreamCompression-16  124MB/s ± 1%  524MB/s ± 2%  +323.25%  (p=0.008 n=5+5)
```

Maybe just add a comment to `SetNbWorkers ` to warn about this behavior ?

bsergean:
Up to you. Maybe we can add a note to the documentation about this odd arm64 behavior, and a link to a ticket ?

bsergean:
Also, maybe that file is special, but the speedup of 33% on your benchmark is a bit disappointing. (Maybe because 8 cores isn’t big enough ?)

For compressing a 4.6 gig file of mine, I almost get a 6x speedup. 

4.63 GiB =>    716 MiB

zstd
8.58s user 1.11s system 113% cpu 8.528 total

zstdmt (multi core)
9.85s user 1.29s system 612% cpu 1.819 total


2. And when using the Mozilla.tar file, it look like a 2x speedup.

zstd$ time zstd mozilla.tar 
mozilla.tar          : 36.56%   (  48.8 MiB =>   17.9 MiB, mozilla.tar.zst)    
zstd mozilla.tar  0.15s user 0.05s system 119% cpu 0.166 total
zstd$ rm mozilla.tar.zst 
zstd$ time zstdmt mozilla.tar
mozilla.tar          : 36.56%   (  48.8 MiB =>   17.9 MiB, mozilla.tar.zst)    
zstdmt mozilla.tar  0.21s user 0.05s system 326% cpu 0.080 total

Viq111:
Sorry, the one above is actually a benchmark on x86 (editing my message to reflect) it:

x86 (`c6i.4xlarge`):
```
name                  old time/op   new time/op   delta
StreamCompression-16    414ms ± 1%     98ms ± 2%   -76.37%  (p=0.008 n=5+5)

name                  old speed     new speed     delta
StreamCompression-16  124MB/s ± 1%  524MB/s ± 2%  +323.25%  (p=0.008 n=5+5)
```
so 3.2x increase (for 8 workers)

ARM (`c7g.4xlarge`):
```
name                  old time/op   new time/op   delta
StreamCompression-16    513ms ± 1%     74ms ± 1%   -85.61%  (p=0.008 n=5+5)

name                  old speed     new speed     delta
StreamCompression-16  100MB/s ± 1%  694MB/s ± 1%  +594.94%  (p=0.008 n=5+5)
```
So ~6x increase (for 8 workers)


So this is not a behavior only on ARM but that calls to `write.Write(...)` actually becomes async.
So maybe add a comment of the sort:
```
// Set the number of workers to run the compression in parallel using multiple threads
// If > 1, the Write() call will become asynchronous. This means data will be buffered until processed.
// If you call Write() too fast, you might incur a memory buffer up to as large as your input.
// Consider calling Flush() periodically if you need to compress a very large file that would not fit all in memory.
```

bsergean:
I just updated the documentation based on your recommendation.

bsergean:
Also it’s interesting that the speedup is bigger on arm cpus ...

bsergean:
I actually use io.Copy in my own code.

        writer := zstd.NewWriter(f)
        w := io.MultiWriter(writer, h)

        if err != nil {
            return -1, "", err
        }
        if _, err = io.Copy(w, reader); err != nil {
            return -1, "", err
        }

        writer.Flush()

        checksum = fmt.Sprintf("%x", h.Sum64())
        writer.Close()

(I’m computing a checksum at the same time)

I wonder how we should update this code to periodically call Flush, if io.Copy will deal with that under the hood, or if I’ll run in the same memory problems as the benchmark.

Viq111:
I would leave the API as is because it gives the user a choice to call Flush whenever. As a headsup, calling `Flush()` makes the payload a bit bigger as you are forcing zstd to flush as much data as possible.
For your code, you need to ensure either `f` is smaller enough to fit in memory (the pathological case of the benchmark is it was generating massive files) OR you can create an intermediate `io.Write` (similar to [LimitedReader](https://pkg.go.dev/io#LimitedReader)) that periodically calls flush but I think that should live outside of the zstd package

Viq111 (approved):
Thanks for bearing with me as we benchmarked!
The PR as it stands looks good to me and if you are fine with the API, we can merge in its current state

bsergean:
Yes the API is fine with me, I think you did most of the hard work here, thanks for reviewing / working on the PR !

PR #115, opened 2022-04-06
[zstd][#60] Add decompression size sanity Check

Viq111:
Fix #60 (see for full description).

Previously, we were trusting the input from the zstd header to allocate the memory on `Decompress` API. This meant a carefully crafted payload could make the zstd wrapper allocates an arbitrary size:

```
ᐅ git checkout eaf4b06330d96d58d043f862d1c25f4bc8db7f37
ᐅ go test -v
[...]
--- PASS: TestLegacy (0.00s)
    --- PASS: TestLegacy/0 (0.00s)
    --- PASS: TestLegacy/1 (0.00s)
=== RUN   TestBadPayloadZipBomb
^C
signal: interrupt
FAIL	github.com/DataDog/zstd	82.757s
```
Screen Shot 2022-04-06 at 17 34 53
Screen Shot 2022-04-06 at 17 35 00

After this fix, we correctly don't allocate too much.
The main part of the change is at: https://github.com/DataDog/zstd/commit/30c4b2986e0ce4fa8ad4b5ee6c1d9d845326d43b#diff-fc46d6feefba07384476b9d2f5a4ae18d6f7322dcd684c388ebe1b2b8d19d394R55-R74

We now basically allocate up to: `min(ZSTD_getFrameContentSize(), max(1MB, 10*input_size))`

Issue #60, opened 2019-04-28
Decoder doesn't check output size before allocating


klauspost:
Please answer these questions before submitting your issue. Thanks!

### What version of Go are you using (`go version`)?

`go version go1.12.1 windows/amd64`

### What operating system and processor architecture are you using (`go env`)?

```
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\mfiles\AppData\Local\go-build
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=e:\gopath\
set GOPROXY=
set GORACE=
set GOROOT=c:\go
set GOTMPDIR=
set GOTOOLDIR=c:\go\pkg\tool\windows_amd64
set GCCGO=gccgo
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -mthreads -fmessage-length=0 -fdebug-prefix-map=e:\temp\wintemp\go-build152196250=/tmp/go-build -gno-record-gcc-switches
```

Package version: `809b919c325d7887bff7bd876162af73db53e878` (v.1.4.0 tag)

### What did you do?

Fuzz test. Using `ref, refErr := zstd.Decompress(nil, data)`.

It seems like the decompressor will attempt to [allocate whatever value is returned](https://github.com/DataDog/zstd/blob/1.x/zstd.go#L130) from `ZSTD_getDecompressedSize`, meaning you can crash the decoder with a very small payload. The value is also cast from uint64 to `int` allowing negative values, which will `panic: runtime error: makeslice: len out of range`. Note that `int` can also be 32 bits.

Sample: [a43e4d2eb6da97e23c23964ee90a093262f69564.zst.gz](https://github.com/DataDog/zstd/files/3125345/a43e4d2eb6da97e23c23964ee90a093262f69564.zst.gz)  (gzipped to make github happy).

### What did you expect to see?

No crash. The decoder should fall back to streaming over a certain size.

### What did you see instead?

```
runtime: VirtualAlloc of 52983526064128 bytes failed with errno=1455
fatal error: out of memory

runtime stack:
[...]

PR #118, opened 2022-05-23
[zstd] remove src minimal size limit on `decompressSizeHint`

XiaochenCui:
We want to mandate the 'zstd' decompressed data to the buffer pre-allocated by us in some circumstances. (No matter whether its size is smaller than `zstdFrameHeaderSizeMax`.)

For example, in our OLAP system kernel, we have a unified buffer pool to manage all bytes allocated for queries. An accidentally created large array will have a fatal impact on the correctness and performance of the system. (In this case, any compressed data which have a size smaller than 18 would result in a newly constructed byte slice with a size of 1MB, which is intolerable.)

I had also considered other workaround:
- provide a global option to control the manner of binary allocation, the pros are it will make our condition branches simpler, and the cons are the difficulty of this library will get increased, and it will be tricky when different components of a system want different strategies.
- provide a new decompression API, which will break the consistence of our library.
- provide a new arg in the old API, same as point 2.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -67,6 +67,10 @@ func decompressSizeHint(src []byte) int {
 	}
 
 	hint := int(C.ZSTD_getFrameContentSize(unsafe.Pointer(&src[0]), C.size_t(len(src))))
+	if hint == 0 {
+		// The minimal hint size is 1, beacuse if we return a size with 0, the following access operation will panic with `index out of range`.
+		hint = 1
+	}
 	if hint < 0 { // On error, just use upperBound
 		hint = upperBound
 	}
--- a/zstd.go
+++ b/zstd.go
@@ -66,12 +66,9 @@ func decompressSizeHint(src []byte) int {
 		upperBound = decompressSizeBufferLimit
 	}
 
-	hint := upperBound
-	if len(src) >= zstdFrameHeaderSizeMax {
-		hint = int(C.ZSTD_getFrameContentSize(unsafe.Pointer(&src[0]), C.size_t(len(src))))
-		if hint < 0 { // On error, just use upperBound
-			hint = upperBound
-		}
+	hint := int(C.ZSTD_getFrameContentSize(unsafe.Pointer(&src[0]), C.size_t(len(src))))
+	if hint < 0 { // On error, just use upperBound
+		hint = upperBound
 	}
 
 	// Take the minimum of both
--- a/zstd.go
+++ b/zstd.go
@@ -37,6 +37,9 @@ const (
 	decompressSizeBufferLimit = 1000 * 1000
 
 	zstdFrameHeaderSizeMax = 18 // From zstd.h. Since it's experimental API, hardcoding it
+
+	// From: https://github.com/DataDog/zstd/blob/1.x/zstd.h#L1123
+	zstdFrameHeaderSizeMin = 2
 )
 
 // CompressBound returns the worst case size needed for a destination buffer,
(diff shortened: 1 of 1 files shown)


XiaochenCui:
Closed it for not passed tests

XiaochenCui:
Bug fixed

XiaochenCui:
@Viq111 
Request review or better solution on this issue sincerely.

Viq111:
Hello @XiaochenCui , thanks for the PR.
Would you have a example payload that would be smaller than 18 bytes and valid zstd ?
Indeed the 18 bytes for the frame is the max, we could change it to the [min with is 2 bytes](https://github.com/DataDog/zstd/blob/1.x/zstd.h#L1123)

XiaochenCui:
> Hello @XiaochenCui , thanks for the PR. Would you have a example payload that would be smaller than 18 bytes and valid zstd ? Indeed the 18 bytes for the frame is the max, we could change it to the [min with is 2 bytes](https://github.com/DataDog/zstd/blob/1.x/zstd.h#L1123)

Yeah, we have a compressed column in our OLAP system with a length of 16. It's generated by a binary with 52 zeros.

Here is the code:
```go
originalData = make([]byte, 52)
logging.Printf("len of original data: %d", len(originalData))

compressedData = Compress(CompressZSTD, originalData)
logging.Printf("len of compressed data: %d, content: %v", len(compressedData), compressedData)
```

The `Compress` function is just a wrapper for `zstd2.CompressLevel(...)`

And here is the output:
```
2022-05-29 16:56:50 | INFO  | compress_test.go:90 > len of original data: 52
2022-05-29 16:56:50 | INFO  | compress_test.go:93 > len of compress data: 16, content: [40 181 47 253 32 52 61 0 0 8 0 1 0 224 116 66]
```

XiaochenCui:
In zstd1.50, the `zstd2.CompressLevel` will always use our buffer for decompressed data.

However, in zstd1.52, it will generate a huge binary with size=1 million when we have compressed data with a size < 18, which is a critical problem for both performance and consistency. We downgrade the library to 1.50 to erase this trouble. But I really want to solve it from the root. (I thought the root cause is `decompressSizeHint`, tell me if it isn't)

XiaochenCui:
Here is the code that is more convenient for you:

```go
        originalData := make([]byte, 52)
	logging.Printf("len of original data: %d", len(originalData))

	// compressedData = Compress(CompressZSTD, originalData)
	compressedData, err := zstd2.CompressLevel(make([]byte, 0, len(originalData)), originalData, zstd2.DefaultCompression)
	if err != nil {
		panic(err)
	}
	logging.Printf("len of compressed data: %d, content: %v", len(compressedData), compressedData)`
```

XiaochenCui:
@Viq111 
I have updated the minimal size of `hint` to `zstdFrameHeaderSizeMin`, which is 2.

Btw, why do we have to "prevent DOS from maliciously-created payloads"? As I'm not good at application cryptography.

Viq111:
Hello!

Ah I've been working on the code in parallel yesterday, here is the PR with a test: https://github.com/DataDog/zstd/pull/120 that should fix your issue (thanks for reporting!)

> Btw, why do we have to "prevent DOS from maliciously-created payloads"? As I'm not good at application cryptography.
See the original PR that introduced the change: #115 (and related issue #60) for the explanation.

If that's ok, I will merge #120 instead of this one as it has the added test ?

XiaochenCui:
> Hello!
> 
> Ah I've been working on the code in parallel yesterday, here is the PR with a test: #120 that should fix your issue (thanks for reporting!)
> 
> > Btw, why do we have to "prevent DOS from maliciously-created payloads"? As I'm not good at application cryptography.
> > See the original PR that introduced the change: #115 (and related issue #60) for the explanation.
> 
> If that's ok, I will merge #120 instead of this one as it has the added test ?

Sure, I've reviewed that pr and it's great! Thanks for tracking this issue.

PR #120, opened 2022-06-02
[zstd] Fix unneededly allocate large decompression buffer

Viq111:
Fix #118
With the introduction of #115 to prevent zipbombs, the check was too strict and was checking a too large size boundary. This fixes it and adds a test

Thanks to @XiaochenCui for reporting

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -36,7 +36,7 @@ const (
 	// decompressed, err := zstd.Decompress(dst, src)
 	decompressSizeBufferLimit = 1000 * 1000
 
-	zstdFrameHeaderSizeMax = 18 // From zstd.h. Since it's experimental API, hardcoding it
+	zstdFrameHeaderSizeMin = 2 // From zstd.h. Since it's experimental API, hardcoding it
 )
 
 // CompressBound returns the worst case size needed for a destination buffer,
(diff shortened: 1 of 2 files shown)


XiaochenCui:
LGTM

zstd.go
```diff
@@ -67,11 +67,14 @@ func decompressSizeHint(src []byte) int {
 	}
 
 	hint := upperBound
-	if len(src) >= zstdFrameHeaderSizeMax {
+	if len(src) >= zstdFrameHeaderSizeMin {
```
david-ds, line 70: why not check the upper bound as well?

```go
len(src) >= zstdFrameHeaderSizeMin && len(src) <= zstdFrameHeaderSizeMax
```

Viq111: I'm not sure what you mean ? That check if just to make sure the header is in the payload but src will 99% of the time be bigger than `zstdFrameHeaderSizeMax` (and that's fine)

david-ds: 👍 sorry I thought that `src` would only be the header, I didn't realize it was the full payload.

david-ds (approved):

Viq111:
Thanks for the report @XiaochenCui ! I'm still looking at #117 and once it's ready, I'll cut a new release

XiaochenCui:
> Thanks for the report @XiaochenCui ! I'm still looking at #117 and once it's ready, I'll cut a new release

Sure thing! It feels great to contribute to the open-source community.

PR #130, opened 2023-06-16
add DecompressInto for decompression of payloads with known sizes

jbowens:
In some systems, compressed payloads are prefixed or tagged with the size of the original payload. This allows allocation of a perfectly-sized buffer to hold the decompressed payload. The current Decompress interface prohibits this type of usage, because Decompress allocates if the destination buffer is smaller than the worst-case bound. In practice, a perfectly-sized buffer will always be smaller than what's required for the worst-case payload.

This commit introduces an additional DecompressInto entrypoint. DecompressInto accepts a destination buffer and requires that the decompressed payload fit into the buffer. If it does not, it returns the original error returned by zstd.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -143,12 +143,7 @@ func Decompress(dst, src []byte) ([]byte, error) {
 		dst = make([]byte, bound)
 	}
 
-	written := int(C.ZSTD_decompress(
-		unsafe.Pointer(&dst[0]),
-		C.size_t(len(dst)),
-		unsafe.Pointer(&src[0]),
-		C.size_t(len(src))))
-	err := getError(written)
+	written, err := DecompressInto(dst, src)
 	if err == nil {
 		return dst[:written], nil
 	}
@@ -160,4 +155,20 @@ func Decompress(dst, src []byte) ([]byte, error) {
 	r := NewReader(bytes.NewReader(src))
 	defer r.Close()
 	return ioutil.ReadAll(r)
+}
+
+// DecompressInto decompresses src into dst. Unlike Decompress, DecompressInto
+// requires that dst be sufficiently large to hold the decompressed payload.
+// DecompressInto may be used when the caller knows the size of the decompressed
+// payload before attempting decompression.
+//
+// It returns the number of bytes copied and an error if any is encountered. If
+// dst is too small, DecompressInto errors.
+func DecompressInto(dst, src []byte) (int, error) {
+	written := int(C.ZSTD_decompress(
+		unsafe.Pointer(&dst[0]),
+		C.size_t(len(dst)),
+		unsafe.Pointer(&src[0]),
+		C.size_t(len(src))))
+	return written, getError(written)
 }
--- a/zstd_test.go
+++ b/zstd_test.go
@@ -87,6 +87,35 @@ func TestCompressDecompress(t *testing.T) {
 	}
 }
 
+func TestCompressDecompressInto(t *testing.T) {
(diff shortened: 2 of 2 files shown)


jbowens:
@Viq111 — Please let me know your thoughts. Happy to rework the interface as needed.

jbowens:
@Viq111 — Is this something that you might be willing to merge upstream, or should I fork?

zstd.go
```diff
@@ -161,3 +161,18 @@ func Decompress(dst, src []byte) ([]byte, error) {
 	defer r.Close()
 	return ioutil.ReadAll(r)
 }
+
+// DecompressInto decompresses src into dst. Unlike Decompress, DecompressInto
```
Viq111, line 165: could you have `Decompress` calls `DecompressInto` at l.146 so we don't have to repeat the call to `C.ZSTD_decompress` ?

zstd.go
```diff
@@ -161,3 +161,18 @@ func Decompress(dst, src []byte) ([]byte, error) {
...
+// requires that dst be sufficiently large to hold the decompressed payload.
+// DecompressInto may be used when the caller knows the size of the decompressed
+// payload before attempting decompression.
+//
+// If dst is too small, DecompressInto returns an error.
+func DecompressInto(dst, src []byte) error {
```
Viq111, line 171: to make it a bit closer to golang stdlib, I'd probably have the method signature also returns the nb of bytes written (`DecompressInto(dst, src []byte) (int, error)` like https://pkg.go.dev/io#Copy). You can always ignore it for your use-case but I think it will usage a bit more wide at no extra cost

Viq111 (commented):
thanks for your contribution!
Sorry for the delay, the end of quarter is always a bit busier.
After the comments are resolved, the PR lgtm!

zstd.go
jbowens: Done.

zstd.go
jbowens: Makes sense, done.

jbowens:
> Sorry for the delay, the end of quarter is always a bit busier.

No worries! Thanks for reviewing :)

Viq111 (approved):
No change on benchmarks:
```
goos: darwin
goarch: arm64
pkg: github.com/DataDog/zstd
                       │     old      │                 new                 │
                       │    sec/op    │    sec/op      vs base              │
Compression-10           66.37m ± ∞ ¹    67.24m ± ∞ ¹       ~ (p=0.421 n=5)
Decompression-10         9.424m ± ∞ ¹    9.336m ± ∞ ¹       ~ (p=0.222 n=5)
geomean                  1.901m          1.916m        +0.84%
¹ need >= 6 samples for confidence interval at level 0.95

                       │      old       │                 new                  │
                       │      B/s       │      B/s        vs base              │
Compression-10            143.3Mi ± ∞ ¹    141.4Mi ± ∞ ¹       ~ (p=0.421 n=5)
Decompression-10         1008.9Mi ± ∞ ¹   1018.5Mi ± ∞ ¹       ~ (p=0.222 n=5)
geomean                   259.7Mi          257.6Mi        -0.83%
¹ need >= 6 samples for confidence interval at level 0.95
```

ribasushi:
@Viq111 @jbowens it seems that the new method has not been added to the `zstd.NewCtx()` object. Is this intentional or an oversight?

jbowens:
@ribasushi — It wasn't intentional on my side; my project does not use the zstd context API and I only had passing awareness of its existence.

PR #139, opened 2024-06-06
Add context functions.

StephenWakely:
The [docs](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L253-L255) recommend allocating a context once and reusing this for each compression:

```
 *  When compressing many times,
 *  it is recommended to allocate a context just once,
 *  and reuse it for each successive compression operation.
```

We have just started using this library in the Datadog agent. Since the agent will be compressing *a lot* I feel it would make sense to expose and use these functions so that we don't have to allocate and free the context with each call.

There is a bunch of code duplication here - I've just copied the existing functions. Let me know if you would prefer to share the code and I will make the changes accordingly.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -84,6 +84,59 @@ func decompressSizeHint(src []byte) int {
 	return hint
 }
 
+// Type alias for the compression context returned by `CreateCCtx`.
+type CCtx = C.ZSTD_CCtx
+
+// Allocates a compression context that can be reused for multiple compression operations.
+func CreateCCtx() *CCtx {
+	return C.ZSTD_createCCtx()
+}
+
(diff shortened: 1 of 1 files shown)


zstd.go
```diff
@@ -84,6 +84,58 @@ func decompressSizeHint(src []byte) int {
 	return hint
 }
 
+// Type alias for the compression context returned by `CreateCCtx`.
+type CCtx = C.ZSTD_CCtx
```
akrylysov, line 88: Do you mind adding tests for the new functions?

StephenWakely:
Ah, it turns out the functions I am adding here are already covered in https://github.com/DataDog/zstd/blob/1.x/zstd_ctx.go.

This isn't needed.

PR #144, opened 2024-09-04
zstd_stream.go: remove intermediate source buffer for Writer.

knusbaum:
The Writer was using an intermediate source buffer, which could sometimes buffer input bytes until the next Write, Flush, or Close call. This is not necessary, complicates the code, and can cause a performance hit.

This commit eliminates the source buffer and its overhead, instead ensuring that all input bytes are compressed when Write returns.

In practice the intermediate source buffer should never or almost never have been used, so this is mostly beneficial for simplifying the code, although we do see a memory usage improvement in the benchmarks

```
name                    old time/op    new time/op    delta
StreamCompression-16      57.9ms ± 5%    56.7ms ± 5%    ~     (p=0.165 n=10+10)

name                    old speed      new speed      delta
StreamCompression-16    1.81GB/s ± 5%  1.85GB/s ± 5%    ~     (p=0.165 n=10+10)

name                    old alloc/op   new alloc/op   delta
StreamCompression-16      2.19MB ± 7%    2.07MB ± 6%  -5.45%  (p=0.001 n=10+10)

name                    old allocs/op  new allocs/op  delta
StreamCompression-16        3.00 ± 0%      3.00 ± 0%    ~     (all equal)
```

### DIFF
--- a/zstd_stream.go
+++ b/zstd_stream.go
@@ -80,7 +80,6 @@ type Writer struct {
 
 	ctx              *C.ZSTD_CCtx
 	dict             []byte
-	srcBuffer        []byte
 	dstBuffer        []byte
 	firstError       error
 	underlyingWriter io.Writer
@@ -138,7 +137,6 @@ func NewWriterLevelDict(w io.Writer, level int, dict []byte) *Writer {
 		CompressionLevel: level,
 		ctx:              ctx,
 		dict:             dict,
-		srcBuffer:        make([]byte, 0),
 		dstBuffer:        make([]byte, CompressBound(1024)),
 		firstError:       err,
 		underlyingWriter: w,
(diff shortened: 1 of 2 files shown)


zstd_stream.go
```diff
@@ -223,24 +216,18 @@ func (w *Writer) Flush() error {
...
-		}
-
 		C.ZSTD_compressStream2_flush(
 			w.resultBuffer,
 			w.ctx,
 			unsafe.Pointer(&w.dstBuffer[0]),
 			C.size_t(len(w.dstBuffer)),
-			unsafe.Pointer(srcPtr),
-			C.size_t(len(w.srcBuffer)),
+			unsafe.Pointer(uintptr(0)),
```
ajgajg1134, line 224: 🤔 does this mean that `flush` now never does anything since we never have any data left in the source buffer?

zstd_stream.go
```diff
@@ -267,24 +254,18 @@ func (w *Writer) Close() error {
...
-		}
-
 		C.ZSTD_compressStream2_finish(
 			w.resultBuffer,
 			w.ctx,
 			unsafe.Pointer(&w.dstBuffer[0]),
 			C.size_t(len(w.dstBuffer)),
-			unsafe.Pointer(srcPtr),
-			C.size_t(len(w.srcBuffer)),
+			unsafe.Pointer(uintptr(0)),
```
ajgajg1134, line 262: same here? If there's never anything left to flush can we simplify this and not call compressStream at all and just return any known error / free any resources needed?

ajgajg1134 (commented):
This definitely looks clearer to me! I wonder if there's a possible latency regression here where if zstd fails to compress all the data in `write` in a single call, previously that extra work would move to when `flush` is called, but now all that work happens in the `write` call? I don't know enough about how this is used to know if that's a realistic possibility though :P

zstd_stream.go
```diff
@@ -154,65 +152,60 @@ func (w *Writer) Write(p []byte) (int, error) {
...
+			w.resultBuffer,
+			w.ctx,
+			unsafe.Pointer(&w.dstBuffer[dstoff]),
+			C.size_t(len(w.dstBuffer[dstoff:])),
+			unsafe.Pointer(&p[0]),
+			C.size_t(len(p)),
+		)
+		ret := int(w.resultBuffer.return_code)
+		if err := getError(ret); err != nil {
+			// TODO: Why do we not set firstError?
```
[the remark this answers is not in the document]
knusbaum: Yes, Sorry I forgot to remove the TODO.
The old code did not do this, and I think that's an error. Tests pass with this change.

zstd_stream.go
knusbaum: Flush can still do something because zstd maintains internal buffers.
i.e. if you write 1 byte in `Write()`, it's likely that zstd will buffer that internally so it can actually do compression on it.

Flush needs to be done to totally clear zstd's internal state.

zstd_stream.go
knusbaum: Nope, same reason as above.

zstd_stream.go
ajgajg1134: Ahhh okay, thanks for the explanation!

ajgajg1134 (approved):

Viq111 (approved):

knusbaum:
Fixed. I'll merge this once CI is done.

PR #150, opened 2025-03-06
Add Reader max window size and Writer window size

### LINKED ISSUE #123: Setting decompression WindowSize?
Please answer these questions before submitting your issue. Thanks!

### What version of Go are you using (`go version`)?

go version go1.20.1 darwin/amd64

### What operating system and processor architecture are you using (`go env`)?

MacOS Monterey Intel i5 64 bit

### What did you do?

I am trying to use NewReader() for decompressing huge zstd archives from Reddit in stream mode, but I am getting
`failed to decompress: Frame requires too much memory for decoding`
I was using https://github.com/klauspost/compress, which has `zstd.WithWindowSize(2<<30)` so it was working fine, but it uses twice the memory so I end up occupying 4gb for decompression instead of 2gb window size as mentioned here:

https://github.com/klauspost/compress/discussions/675#discussioncomment-3736007

### What did you expect to see?

I was hoping your lib can expose this window size option so I can see if memory usage decreases?

### What did you see instead?

I am getting error `failed to decompress: Frame requires too much memory for decoding`

cce:
Adds `NewWriterLevelDictWindowSize` and `NewReaderDictMaxWindowSize` to limit memory usage during compression and decompression. 

- NewWriterLevelDictWindowSize added to set specific window size (in bytes, not as log base 2)
- NewReaderDictMaxWindowSize added to set a maximum window size that this decompressor will allow, also in bytes.
- Tests exercising the Writer can set window sizes, and that limiting them in the Reader works.

More documentation on window sizes are in [RFC 9659](https://www.rfc-editor.org/rfc/rfc9659.html) and in zstd.h:
```
/*! ZSTD_DCtx_setMaxWindowSize() :
 *  Refuses allocating internal buffers for frames requiring a window size larger than provided limit.
 *  This protects a decoder context from reserving too much memory for itself (potential attack scenario).
 *  This parameter is only useful in streaming mode, since no internal buffer is allocated in single-pass mode.
 *  By default, a decompression context accepts all window sizes <= (1 << ZSTD_WINDOWLOG_LIMIT_DEFAULT)
 * @return : 0, or an error code (which can be tested using ZSTD_isError()).
 */
```
```
    ZSTD_c_windowLog=101,    /* Maximum allowed back-reference distance, expressed as power of 2.
                              * This will set a memory budget for streaming decompression,
                              * with larger values requiring more memory
                              * and typically compressing more.
                              * Must be clamped between ZSTD_WINDOWLOG_MIN and ZSTD_WINDOWLOG_MAX.
                              * Special: value 0 means "use default windowLog".
```
Resolves #123.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Decompress(dst, src []byte) ([]byte, error)
 NewWriter(w io.Writer) *Writer
 NewWriterLevel(w io.Writer, level int) *Writer
 NewWriterLevelDict(w io.Writer, level int, dict []byte) *Writer
+NewWriterLevelDictWindowSize(w io.Writer, level int, dict []byte, windowSize int) *Writer
 
 // Write compresses the input data and write it to the underlying writer
 (w *Writer) Write(p []byte) (int, error)
@@ -89,6 +90,7 @@ NewWriterLevelDict(w io.Writer, level int, dict []byte) *Writer
 // to call Close, which frees up C objects.
 NewReader(r io.Reader) io.ReadCloser
 NewReaderDict(r io.Reader, dict []byte) io.ReadCloser
+NewReaderDictMaxWindowSize(r io.Reader, dict []byte, maxWindowSize int) io.ReadCloser
 ```
 
 ### Benchmarks (benchmarked with v0.5.0)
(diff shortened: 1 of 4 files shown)


Viq111 (commented):
Thanks for your contribution and the thorough test cases !

Unfortunately as this is using experimental (per zstd's documentation) features and disallow dynamic linking, we wouldn't be able to merge to the `1.x`. I'll leave the PR open and if the upstream zstd repositority makes this parameter non-experimental, we can revisit and merge it

PR #151, opened 2025-03-18
add Ctx.DecompressInto for decompression of payloads with known sizes

EdwardX29:
Adds DecompressInto() to zstd_context to match the functionality previously implemented in the zstd file.

In some systems, compressed payloads are prefixed or tagged with the size of the original payload. This allows allocation of a perfectly-sized buffer to hold the decompressed payload. This method provides a way to decompress directly into a pre-allocated buffer.

DecompressInto accepts a destination buffer and requires that the decompressed payload fit into the buffer. If it does not, it returns the original error returned by zstd.

Relates to/Follows up on: https://github.com/DataDog/zstd/pull/130

### DIFF
--- a/zstd_ctx.go
+++ b/zstd_ctx.go
@@ -24,6 +24,15 @@ type Ctx interface {
 	// prevent allocation.  If it is too small, or if nil is passed, a new buffer
 	// will be allocated and returned.
 	Decompress(dst, src []byte) ([]byte, error)
+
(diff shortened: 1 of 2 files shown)


jbowens:
@viq111 tagging you since you reviewed #130. Is this something that we could upstream?

Viq111 (approved):
Sorry for the delay, this looks great! Thanks for your contribution!

PR #154, opened 2025-07-10
fix decompressSizeHint logic error

zjj:


Viq111 (changes requested):
I'm so sorry, I read your change incorrectly and the existing behavior is the correct one. I only realized it after seeing the tests.

 I guess it needs an explanation and I'll add a comment in that direction:
- Most of the time, the payload is genuine and thus `hint = getFrameContentSize(src)` is correct
- However when decompressing from an untrusted source (e.g: an intake), we want to make sure that payload can only use a "fair" amount of memory. Fair being defined are `max(smallFixedSize, O(size of input))`. Currently `smallFixedSize = 1 MB` and `O` is 50x.
  - For "small" payload, it is ok that the decompress buffer can go up to 1MB (but is still `min(zstdHintFromSrc, 1MB)`
  - For large payload, we want to follow that linear size related to the input size

Hence the current behavior.

If we were to change it, this would break decompress for any payload where size of output > 1MB (if `Decompress` is called with `dst == nil`)

PR #155, opened 2025-08-26
Keep decompressing as long as we have compressed data

vickenty:
ZSTD_decompressStream may consume only part of the src buffer, for example if the input is a concatenation of two compressed streams. In this case the second stream will remain undecoded in the input buffer after the first call, and we need to call decompress routine again.

This makes NewReader behavior closer to Decompress, that handles such concatenated streams transparently.

### DIFF
--- a/zstd_stream.go
+++ b/zstd_stream.go
@@ -355,6 +355,7 @@ type reader struct {
 	decompressionBuffer []byte
 	decompOff           int
 	decompSize          int
+	remainingBytes      int
 	dict                []byte
 	firstError          error
 	recommendedSrcSize  int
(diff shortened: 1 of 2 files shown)


zstd_stream.go
```diff
@@ -447,12 +448,12 @@ func (r *reader) Read(p []byte) (int, error) {
...
-		// - If the last decompression did entirely fill the decompression buffer,
-		//   it might have needed more room to decompress the input. In that case,
-		//   don't do any unnecessary Read that might block.
-		needsData := r.decompSize < len(r.decompressionBuffer)
+		// Do a read only when zstd doesn't have any pending output. When zstd could
+		// not finish writing output, it will keep input buffer non-empty.
+		// https://github.com/facebook/zstd/commit/b3060f7a9ea3555c6045606be58dddc86bbb099b
+		// > when all compressed frame is consumed, it means decompression is completed,
+		// > with regenerated data fully flushed.
+		needsData := r.compressionLeft == 0
```
Viq111, line 456: I was worried about this change potentially inducing an infinite loop (zstd waiting for addl data but consuming 0 from the input buffer), however I did run fuzzing on this endpoint and did not find anything so g2g

Viq111 (approved):

PR #163, opened 2026-04-20
Fix C memory leak in Writer.Close() and reader.Close()

### LINKED ISSUE #157: Memory Leak in ZSTD_compressStream2_wrapper
Running tests with AddressSanitizer (ASan) enabled reveals a memory leak when using the zstd.Writer or a streaming compression method. The leak appears to be within the underlying C-allocated context which is not being properly freed by the Go wrapper's cleanup logic, even when the Close() method is explicitly called.

The leak is tracked back to a malloc call within the core Zstandard C library, typically associated with the stream's internal compression context (ZSTD_CCtx*).

eric-weaver:
## Fix C memory leak in `Writer.Close()` and `reader.Close()`

Fixes #157

### The Bug

`Writer.Close()` returns early when `firstError` is set **without calling `ZSTD_freeCStream`**, permanently leaking the C-allocated compression context:

```go
// Before (zstd_stream.go)
func (w *Writer) Close() error {
    if w.firstError != nil {
        return w.firstError // ← LEAK: ctx never freed
    }
    // ...
    return getError(int(C.ZSTD_freeCStream(w.ctx))) // only reached if no error
}
```

A secondary leak exists when `getError(ret)` fails during the flush loop — it returns without freeing. The one path that *does* free (underlying writer error) never nils `ctx`, causing **use-after-free on double-close**.

`reader.Close()` has the same early-return leak with `ZSTD_freeDStream`.

### The Fix

Three changes to `zstd_stream.go`:

1. **Free `ctx` on every exit path and nil it** — both `Writer.Close()` and `reader.Close()` now call `ZSTD_freeCStream`/`ZSTD_freeDStream` and set `ctx = nil` regardless of which error path is taken.

2. **Nil guard for double-close** — `Close()` checks `ctx == nil` at entry and returns immediately, preventing use-after-free if called more than once.

3. **`runtime.SetFinalizer` safety net** — `NewWriterLevelDict` and `NewReaderDict` now register a finalizer to free the C context if the caller forgets to `Close()`, matching the existing pattern used by `finalizeCtx` and `finalizeBulkProcessor`.

### Valgrind Verification

Ran a 100k-iteration leak test under valgrind via Docker:

```go
func TestWriterCloseMemoryLeakLoop(t *testing.T) {
    for i := 0; i < 100_000; i++ {
        var buf bytes.Buffer
        w := NewWriter(&buf)
        w.firstError = errors.New("simulated error")
        w.Close()
    }
}
```

```bash
docker run --rm zstd-valgrind \
    -test.run TestWriterCloseMemoryLeakLoop -test.v -test.count=1
```

| Metric | Before | After |
|---|---|---|
| in use at exit | 528,802,016 bytes (100,007 blocks) | 1,728 bytes (6 blocks) |
| allocs / frees | 100,027 / **20** | 100,025 / **100,019** |
| definitely lost | **510,297,288 bytes (96,501 blocks)** | **0 bytes (0 blocks)** |
| possibly lost | 23,168 bytes (11 blocks) | 1,728 bytes (6 blocks) |

The remaining 1,728 bytes "possibly lost" are Go runtime/pthread thread-local storage, unrelated to this library.

### Tests Added

- `TestWriterCloseFreesContextOnFirstError` — ctx is freed when `firstError` is set
- `TestWriterDoubleCloseIsSafe` — second `Close()` is a safe no-op
- `TestWriterCloseNilsContextOnUnderlyingWriterError` — ctx is nil'd after underlying writer error
- `TestReaderCloseFreesContextOnFirstError` — reader ctx is freed when `firstError` is set
- `TestWriterCloseMemoryLeakLoop` — exercises the leak path at scale (valgrind target)

### DIFF
--- a/zstd_stream.go
+++ b/zstd_stream.go
@@ -133,7 +133,7 @@ func NewWriterLevelDict(w io.Writer, level int, dict []byte) *Writer {
 		err = getError(int(C.ZSTD_CCtx_setParameter(ctx, C.ZSTD_c_compressionLevel, C.int(level))))
 	}
 
-	return &Writer{
+	writer := &Writer{
 		CompressionLevel: level,
 		ctx:              ctx,
 		dict:             dict,
@@ -141,6 +141,14 @@ func NewWriterLevelDict(w io.Writer, level int, dict []byte) *Writer {
 		firstError:       err,
 		underlyingWriter: w,
 		resultBuffer:     new(C.compressStream2_result),
+	}
+	runtime.SetFinalizer(writer, finalizeWriter)
+	return writer
+}
+
+func finalizeWriter(w *Writer) {
+	if w.ctx != nil {
+		C.ZSTD_freeCStream(w.ctx)
 	}
 }
 
@@ -247,7 +255,16 @@ func (w *Writer) Flush() error {
 // Close closes the Writer, flushing any unwritten data to the underlying
 // io.Writer and freeing objects, but does not close the underlying io.Writer.
 func (w *Writer) Close() error {
+	if w.ctx == nil {
+		if w.firstError != nil {
+			return w.firstError
+		}
+		return nil
+	}
+
 	if w.firstError != nil {
+		C.ZSTD_freeCStream(w.ctx)
+		w.ctx = nil
 		return w.firstError
 	}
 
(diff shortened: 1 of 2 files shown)


Viq111 (approved):
Thanks for the contribution, lgtm!

PR #168, opened 2026-07-24
Decompress: reuse the caller-supplied buffer for unknown-size frames

KrisKennawayDD:
## Problem

`Decompress` sizes its destination buffer from `decompressSizeHint(src)`, which reads the frame's content-size field. Frames that don't carry that field make the hint fall back to a pessimistic upper bound (`max(50×len(src), decompressSizeBufferLimit)` — at least 1 MB). This happens for:

- **legacy zstd v0.5 frames** (`ZSTD_getFrameContentSize` reports unknown), and
- **streaming frames** compressed without a pledged source size.

When a caller passes a smaller-but-adequate buffer, `Decompress` **discards it** (`cap(dst) < bound`) and allocates the full bound. So every decode of such a frame allocates at least 1 MB regardless of the real payload size — a significant per-call heap/GC cost for callers decoding many small unknown-size payloads with a pooled/hinted buffer.

## Fix

`decompressSizeHint` now also returns `foundHint` — whether the decompressed size was actually read from the frame header.

`Decompress` and `ctx.Decompress` select the destination buffer with a single switch, make one `DecompressInto`, and fall back to the streaming reader only if that buffer is too small:

- `cap(dst) >= hint` — reuse the caller buffer (unchanged from before).
- `!foundHint && cap(dst) > 0` — size unknown, so reuse the caller buffer as-is rather than allocating the pessimistic bound.
- otherwise — allocate `hint` (the exact size when known; the bound when unknown and no caller buffer was supplied).

Consequences:

- **Known-size frames** (the common case where the same zstd version compressed and decompressed): behavior is unchanged, and a caller passing a too-small buffer does **not** pay for a failed `DecompressInto`.
- **Unknown-size frames with an adequate caller buffer**: no more ≥1 MB allocation — the caller buffer is used.
- **Unknown-size frames with a too-small (or no) caller buffer**: fall back to the streaming reader instead of allocating the ≥1 MB bound.

The same change is applied to `ctx.Decompress`; `BulkProcessor.Decompress` is adapted to the new signature with its behavior preserved.

Regression tests cover caller-buffer reuse (unknown- and known-size), the too-small fallback, `nil` dst, and the `foundHint` signal — run as subtests against both `Decompress` and `ctx.Decompress`.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -57,31 +57,33 @@ func cCompressBound(srcSize int) int {
 	return int(C.ZSTD_compressBound(C.size_t(srcSize)))
 }
 
-// decompressSizeHint tries to give a hint on how much of the output buffer size we should have
-// based on zstd frame descriptors. To prevent DOS from maliciously-created payloads, limit the size
-func decompressSizeHint(src []byte) int {
+// decompressSizeHint returns a suggested output size from the frame header, capped to guard against
+// zip bombs. foundHint is false when the frame does not advertise its size (legacy v0.5 or unpledged
+// streaming frames); the returned hint is then only a pessimistic upper bound.
+func decompressSizeHint(src []byte) (hint int, foundHint bool) {
 	// 1 MB or 50x input size
 	upperBound := 50 * len(src)
 	if upperBound < decompressSizeBufferLimit {
 		upperBound = decompressSizeBufferLimit
 	}
 
-	hint := upperBound
+	hint = upperBound
 	if len(src) >= zstdFrameHeaderSizeMin {
-		hint = int(C.ZSTD_getFrameContentSize(unsafe.Pointer(&src[0]), C.size_t(len(src))))
-		if hint < 0 { // On error, just use upperBound
-			hint = upperBound
-		}
-		if hint == 0 { // When compressing the empty slice, we need an output of at least 1 to pass down to the C lib
-			hint = 1
+		contentSize := int(C.ZSTD_getFrameContentSize(unsafe.Pointer(&src[0]), C.size_t(len(src))))
+		if contentSize >= 0 { // a negative value means the size is unknown or the header is in error
+			foundHint = true
+			hint = contentSize
+			if hint == 0 { // When compressing the empty slice, we need an output of at least 1 to pass down to the C lib
+				hint = 1
+			}
 		}
 	}
 
 	// Take the minimum of both
 	if hint > upperBound {
-		return upperBound
+		return upperBound, foundHint
 	}
-	return hint
+	return hint, foundHint
 }
 
 // Compress src into dst.  If you have a buffer to use, you can pass it to
@@ -131,16 +133,26 @@ func CompressLevel(dst, src []byte, level int) ([]byte, error) {
 // Decompress src into dst.  If you have a buffer to use, you can pass it to
 // prevent allocation.  If it is too small, or if nil is passed, a new buffer
 // will be allocated and returned.
+//
+// Note: for frames that do not advertise their size (legacy v0.5 or unpledged
+// streaming frames) dst may be partially overwritten even if a new slice is
+// returned; do not rely on dst's contents after such a call.
 func Decompress(dst, src []byte) ([]byte, error) {
 	if len(src) == 0 {
 		return []byte{}, ErrEmptySlice
 	}
 
-	bound := decompressSizeHint(src)
-	if cap(dst) >= bound {
-		dst = dst[0:cap(dst)]
-	} else {
-		dst = make([]byte, bound)
+	hint, foundHint := decompressSizeHint(src)
+
+	// Reuse the caller buffer when it is large enough, or when the size is
+	// unknown (the hint is then only an upper bound); otherwise allocate the hint.
+	switch {
+	case cap(dst) >= hint:
+		dst = dst[:cap(dst)]
+	case !foundHint && cap(dst) > 0:
+		dst = dst[:cap(dst)]
+	default:
+		dst = make([]byte, hint)
 	}
 
 	written, err := DecompressInto(dst, src)
--- a/zstd_bulk.go
+++ b/zstd_bulk.go
@@ -110,7 +110,11 @@ func (p *BulkProcessor) Decompress(dst, src []byte) ([]byte, error) {
 		return nil, ErrEmptySlice
 	}
 
-	contentSize := decompressSizeHint(src)
+	// Unlike Decompress, this always sizes from the hint and does not reuse a
+	// too-small caller buffer for unknown-size frames: there is no streaming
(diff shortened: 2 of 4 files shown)


zstd.go
```diff
@@ -136,11 +140,31 @@ func Decompress(dst, src []byte) ([]byte, error) {
 		return []byte{}, ErrEmptySlice
 	}
 
-	bound := decompressSizeHint(src)
-	if cap(dst) >= bound {
+	hint, foundHint := decompressSizeHint(src)
+
+	// When the frame does not advertise its decompressed size, the hint is only
```
Viq111: claude tends to be too verbose on comments btw, could we make it more succinct ?

zstd.go
```diff
@@ -136,11 +140,31 @@ func Decompress(dst, src []byte) ([]byte, error) {
...
+	// a pessimistic upper bound (>= decompressSizeBufferLimit). This happens for
+	// legacy zstd v0.5 frames and for streaming frames compressed without a
+	// pledged source size. Rather than discard an adequate caller-supplied
+	// buffer and allocate that bound, try the caller buffer as-is first.
+	// DecompressInto reports a too-small buffer without writing past it, so this
+	// attempt is safe. When the size is known (the common case where the same
+	// zstd version compressed and decompressed) we skip this and size from the
+	// hint, so callers passing a too-small buffer do not pay for a failed attempt.
+	if !foundHint && cap(dst) > 0 {
+		written, err := DecompressInto(dst[:cap(dst)], src)
```
Viq111: instead of having another new if/logic path, I would just have this part create (or not) a new dst buffer and then let the rest of the code flow (on ErrTooSmall, it auto falls back to stream decompress)

zstd_bulk.go
```diff
@@ -110,7 +110,7 @@ func (p *BulkProcessor) Decompress(dst, src []byte) ([]byte, error) {
 		return nil, ErrEmptySlice
 	}
 
-	contentSize := decompressSizeHint(src)
+	contentSize, _ := decompressSizeHint(src)
```
Viq111, line 117: we should make it the same logic here to be consistent ?

zstd_reuse_buffer_test.go
```diff
@@ -0,0 +1,212 @@
...
+type namedDecompressor struct {
+	name string
+	fn   func(dst, src []byte) ([]byte, error)
+}
+
+// decompressors are the entry points that share the caller-buffer-reuse logic.
+func decompressors() []namedDecompressor {
+	return []namedDecompressor{
+		{"Decompress", Decompress},
+		{"ctx.Decompress", NewCtx().Decompress},
```
Viq111: I think this reuse the same NewCtx() object between tests right ? you might want to instead do a func that create a new object for each test (or call)

zstd_reuse_buffer_test.go
```diff
@@ -0,0 +1,212 @@
...
+		t.Run(d.name, func(t *testing.T) {
+			buf := make([]byte, 8192) // adequate for the payload, far below the bound
+			out, err := d.fn(buf, frame)
+			if err != nil {
+				t.Fatalf("decompress: %v", err)
+			}
+			if !bytes.Equal(out, payload) {
+				t.Fatalf("round-trip mismatch")
+			}
+			if cap(out) != cap(buf) {
```
Viq111: I think you even want stronger where the ptr are the same (i.e: same buffer reused)

zstd_reuse_buffer_test.go
```diff
@@ -0,0 +1,212 @@
...
+				t.Fatalf("caller buffer (cap %d) should be reused, got cap %d", cap(buf), cap(out))
+			}
+		})
+	}
+}
+
+// TestDecompressSizeHintFound verifies foundHint distinguishes frames that
+// advertise their content size (produced by Compress) from those that do not
+// (streaming, as here, and legacy zstd v0.5 frames).
+func TestDecompressSizeHintFound(t *testing.T) {
```
Viq111: imo, this isn't a contract we are giving our user, this is an internal contract of (C++) zstd so we shouldn't have a test for it. i.e: if zstd decides to remove size hint in future versions, it doesn't break our code

zstd_reuse_buffer_test.go
```diff
@@ -0,0 +1,212 @@
...
+// TestDecompressUnknownSizeMultiBlockTooSmall exercises an unknown-size,
+// multi-block frame decoded into a too-small caller buffer: zstd writes some
+// blocks into that buffer before reporting it is too small, so decompression
+// falls back to the streaming API. The returned data must still be correct
+// (the caller buffer may be left partially overwritten; see Decompress).
+func TestDecompressUnknownSizeMultiBlockTooSmall(t *testing.T) {
+	// Larger than the 128 KB block size so the frame spans multiple blocks.
+	payload := bytes.Repeat([]byte("datadog-multiblock-"), 30000) // ~570 KB
+	frame := unknownSizeFrame(t, payload)
+	if _, found := decompressSizeHint(frame); found {
```
Viq111: could we do that test within the `unknownSizeFrame` ? So we don't have to do the check at every test

zstd_reuse_buffer_test.go
```diff
@@ -0,0 +1,212 @@
...
+		t.Fatalf("streaming frame should not advertise its content size")
+	}
+}
+
+// TestDecompressUnknownSizeMultiBlockTooSmall exercises an unknown-size,
+// multi-block frame decoded into a too-small caller buffer: zstd writes some
+// blocks into that buffer before reporting it is too small, so decompression
+// falls back to the streaming API. The returned data must still be correct
+// (the caller buffer may be left partially overwritten; see Decompress).
+func TestDecompressUnknownSizeMultiBlockTooSmall(t *testing.T) {
```
Viq111: imo this test is redundant to previous ones and the stream path is already tested in `zstd_test.go`

zstd_reuse_buffer_test.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+		})
+	}
+}
+
+// TestDecompressKnownSizeCappedHint covers a frame that advertises a size larger
+// than decompressSizeHint's upper bound (so the hint is capped while foundHint
+// stays true): the buffer is sized to the capped bound, DecompressInto reports
+// it too small, and decompression falls back to the streaming API.
+func TestDecompressKnownSizeCappedHint(t *testing.T) {
```
Viq111: should already be tested by other tests right ?

zstd_bulk.go
KrisKennawayDD: Added a comment about why this diverges

Viq111 (approved):
✅ 
(I've disabled circleCI for future runs / PR since we moved to github actions)

PR #169, opened 2026-07-24
Decompress: grow from a small buffer for unknown-size frames with no dst

KrisKennawayDD:
> **Stacked on #168** (base branch `kris.kennaway/decompress-reuse-caller-buffer`). Review/merge #168 first; this PR's diff is just the final commit. It's split out so it can be accepted or rejected independently.

## Problem

For frames that do **not** advertise their decompressed size — legacy zstd v0.5 frames, and streaming frames compressed without a pledged source size — `decompressSizeHint` returns only a pessimistic upper bound (`max(50×len(src), decompressSizeBufferLimit)` = at least 1 MB). #168 fixed the case where the caller passes a reusable buffer. This PR handles the remaining case: **`Decompress(nil, src)` / no caller buffer**, where the code still allocates that ≥1 MB bound up front on every such decode regardless of the real payload size.

## Fix

When the size is unknown and no caller buffer is supplied, grow from `3×len(src)` (`3× → 6× → 12×`) before falling back to the streaming reader, instead of allocating the bound. The size-hint fast path (frames that advertise their size) is unchanged.

## History note

This restores the sizing used before **489b911 "Remove Decompress multiple retries."** That commit dropped the grow loop on the premise that *"zstd 1.3.0+ frames always carry a size hint."* That premise does not hold for legacy v0.5 frames or unpledged streaming frames — exactly the case handled here — so the retries weren't evaluated against it. This reintroduces the grow schedule **only** for the unknown-size + no-buffer path, leaving the 1.3.0+ hint optimization intact.

No DoS regression vs **30c4b29 "Add a sanity limit"**: `growDecompress` never allocates based on an attacker-claimed size (only from the input length, bounded to ≤12×`len(src)`), so it does not reintroduce the unbounded-allocation vector that `decompressSizeBufferLimit` guards. (The streaming fallback's `ReadAll` is unbounded, but that already exists on the current code path.)

Applies to both `Decompress` and `ctx.Decompress`; regression tests (nil-buffer unknown-size avoids the bound; multi-block stream fallback; known-size still exact) run as subtests against both.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -145,14 +145,18 @@ func Decompress(dst, src []byte) ([]byte, error) {
 	hint, foundHint := decompressSizeHint(src)
 
 	// Reuse the caller buffer when it is large enough, or when the size is
-	// unknown (the hint is then only an upper bound); otherwise allocate the hint.
+	// unknown (the hint is then only an upper bound). Otherwise size from the
+	// hint, or grow from a small buffer when the size is unknown and none was
+	// supplied (see growDecompress).
 	switch {
 	case cap(dst) >= hint:
 		dst = dst[:cap(dst)]
 	case !foundHint && cap(dst) > 0:
 		dst = dst[:cap(dst)]
-	default:
+	case foundHint:
 		dst = make([]byte, hint)
+	default:
+		return growDecompress(src, DecompressInto)
 	}
 
 	written, err := DecompressInto(dst, src)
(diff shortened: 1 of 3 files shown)


KrisKennawayDD:
Superseded by the decompressSizeHint approach in the new PR (keeps a single decode path per review feedback rather than adding a retry branch).

